PR #11, opened 2019-03-15
TLS API updates

justinboswell:
These are required to ingest the latest MQTT updates in aws-c-mqtt


aws-common-runtime/CMakeLists.txt
```diff
@@ -27,11 +27,11 @@ if (UNIX AND NOT APPLE)
 endif()
 
 set(AWS_C_IO_URL "https://github.com/awslabs/aws-c-io.git")
-set(AWS_C_IO_SHA "v0.2.2")
+set(AWS_C_IO_SHA "c109ffe3156f759723e0609b46182201c65a64e4")
```
JonathanHenson: why are we not running off of tags here? I'm assuming it's just because we haven't cut them yet.

source/io/TlsOptions.cpp
```diff
@@ -27,25 +27,29 @@ namespace Aws
             TlsContextOptions TlsContextOptions::InitDefaultClient() noexcept
             {
                 TlsContextOptions ctxOptions;
-                aws_tls_ctx_options_init_default_client(&ctxOptions.m_options);
+                aws_tls_ctx_options_init_default_client(&ctxOptions.m_options, aws_default_allocator());
```
JonathanHenson: This needs to honor the configured allocator.

aws-common-runtime/CMakeLists.txt
justinboswell: Correct, I need to update IO, then I'll get back on a version tag

source/io/TlsOptions.cpp
justinboswell: Fixed.

justinboswell:
> I'm assuming we're going to come back and add the byte buffer support stuff in a different review?

Yes. This is just enough to get a fix out to the customer.

PR #15, opened 2019-03-27
Http Client, added URI class, various memory leak fixes.

JonathanHenson:


aws-common-runtime/cmake/BuildAwsCHttp.cmake
```diff
@@ -0,0 +1,56 @@
...
+            -DCMAKE_INSTALL_PREFIX=${AWS_DEPS_INSTALL_DIR}
+            -DCMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}
+            -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+            -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
+            -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY=${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+            -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
+            )
+else()
+    ExternalProject_Add(AwsCHttp
```
justinboswell: We need to pass ENABLE_SANITIZERS into external project builds. Don't have to fix that here, but we should make a pass on that.

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+             * Called as part of the outgoing http message's body (request in client mode, and response in server mode).
+             * `buffer` contains the buffer for you to write into. Keep in mind, that `buffer` may already have
+             * pending data in it, so always append after buffer.buf + buffer.len. You can write up to buffer.capacity
+             * - buffer.len into the buffer.
+             *
+             * You need not write anything to the buffer to keep the stream alive. As long as you return:
+             * AWS_HTTP_OUTGOING_BODY_IN_PROGRESS, you will continue to receive this callback until you
+             * return AWS_HTTP_OUTGOING_BODY_DONE.
+             *
+             * This parameter can be unset on the HttpStream object if you will not be sending a body.
```
justinboswell: s/unset/left empty/, as written it implies that you have to clear it

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+                 *
+                 * A helpful utility for setting this value can be found in Aws::Crt::Io::Uri
+                 *
+                 * this value is copied internally.
+                 */
+                ByteCursor uri;
+                /**
+                 * Array of headers to use for the Request. These values will be copied internally.
+                 */
+                aws_http_header *headerArray;
```
justinboswell: Should we type alias aws_http_header for consistency?

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+                    Allocator *allocator = DefaultAllocator()) noexcept;
+
+                /**
+                 * Establish a new Https Connection to hostName:port, using `socketOptions` for tcp options and
+                 * `tlsConnOptions` for TLS/SSL options.
+                 *
+                 * returns true on success, and false on failure. If false is returned, `onConnectionSetup` will not
+                 * be invoked.
+                 */
+                bool NewConnection(
```
justinboswell: Given that it doesn't return like the rest of the factory patterns, perhaps this would better be named StartConnection? BeginConnection? CreateConnection?

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+                 * returns true on success, and false on failure. If false is returned, `onConnectionSetup` will not
+                 * be invoked.
+                 */
+                bool NewConnection(const ByteCursor &hostName, uint16_t port, const Io::SocketOptions &socketOptions)
+                    const noexcept;
+
+                /**
+                 * See `OnConnectionSetup` for more info. This value cannot be empty.
+                 */
+                OnConnectionSetup onConnectionSetup;
```
justinboswell: For all these values that cannot be empty, should we make them ctor params (even if that ctor is an overload)? I feel like otherwise we're setting the user up for the WINAPI problem of EINVALID_XXX, Lassie-style debugging/appeasement of the API. I hate to suggest a builder, but this use case is something it's good at.

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+                ByteCursor method;
+                /**
+                 * Usually the Path and Query portion of the request uri (assuming the host header has been set).
+                 * There's not validation on this, it can also be the absolute uri if you want/need that.
+                 *
+                 * A helpful utility for setting this value can be found in Aws::Crt::Io::Uri
+                 *
+                 * this value is copied internally.
+                 */
+                ByteCursor uri;
```
justinboswell: Should this be a Uri? Especially since it's optional what the contents are?

source/http/HttpConnection.cpp
```diff
@@ -0,0 +1,363 @@
...
+            void HttpClient::s_onClientConnectionSetup(
+                struct aws_http_connection *connection,
+                int errorCode,
+                void *user_data) noexcept
+            {
+                /**
+                 * Allocate an HttpConnection and seat it to `ConnectionWrapper`'s shared_ptr.
+                 */
+                auto *connectionWrapper = reinterpret_cast<ConnectionWrapper *>(user_data);
+                int retError = errorCode;
```
justinboswell: Why is this copied? Seems like errorCode can fill the job just fine?

source/http/HttpConnection.cpp
```diff
@@ -0,0 +1,363 @@
...
+                struct aws_http_connection *connection,
+                int errorCode,
+                void *user_data) noexcept
+            {
+                (void)connection;
+                /* now that we're shutting down, we can release the internal ref count. */
+                auto *connectionWrapper = reinterpret_cast<ConnectionWrapper *>(user_data);
+                connectionWrapper->client->onConnectionShutdown(connectionWrapper->connection, errorCode);
+                connectionWrapper->connection.reset();
+                connectionWrapper->~ConnectionWrapper();
```
justinboswell: This pattern is so common in this CRT, should we do:
```cpp
template <typename T>
void Delete(T* t, Allocator* allocator)
{
    t->~T();
    aws_mem_release(allocator, t);
}
```
and then while we're at it
```cpp
template <typename T, typename ...Args>
T* New(Allocator* allocator, Args... args)
{
    T* t = reinterpret_cast<T*>(aws_mem_acquire(allocator, sizeof(T)));
    if (!t) return nullptr;
    return new(t) T(std::forward<Args...>(args...));
}

File: source/http/HttpConnection.cpp [hunk 9]
```diff
@@ -0,0 +1,363 @@
...
+            HttpConnection::~HttpConnection()
+            {
+                if (m_connection)
+                {
+                    aws_http_connection_release(m_connection);
+                    m_connection = nullptr;
+                }
+            }
+
+            struct StreamWrapper
```
justinboswell: The MQTT API interchangeably names these Wrapper or CallbackData or UserData. Let's pick one and standardize on it. I personally like UserData or CallbackData, since that's what it is.

justinboswell:
Can you also grab the fixes from: https://github.com/awslabs/aws-crt-cpp/commit/1b5ed8b961542c04721631bb7f7247e9af341720 to fix the MQTT leak while you're at it?

marcomagdy:
What's the reason for using reinterpret_cast everywhere?
If the reasoning is to silence the compiler's type-system, then a C-style cast will do. Using reinterpret_cast means you _want_ the type system's help. And I see that misused in this PR.

The rules are simple:
- going from a typed pointer to `void*` does not require a cast as in `void *p = &number;`
- going from a `void*` to a typed pointer requires a `static_cast`
- casting between related types (by signed-ness or inheritance) requires a `static_cast`
- casting between completely unrelated pointer types requires a `reinterpret_cast`

include/aws/crt/io/TlsOptions.h
```diff
@@ -102,6 +107,52 @@ namespace Aws
...
+                 * alpnList is copied.
+                 *
+                 * returns true if the copy succeeded, or false otherwise.
+                 */
+                bool SetAlpnList(const char *alpnList) noexcept;
+
+                operator bool() const noexcept { return m_isInit; }
+                int LastError() const noexcept { return m_lastError; }
+
+                const aws_tls_connection_options *GetUnderlyingHandle() const noexcept
```
marcomagdy: probably unpopular suggestion:
But this is one case where a conversion operator would yield more elegant code.
```cpp
explicit operator const aws_tls_connection_options*() const
{
    return &m_tls_connection_options;
}
```

include/aws/crt/io/Uri.h
```diff
@@ -0,0 +1,102 @@
...
+                 * builds a URI from `builderOptions`. Upon failure the bool() operator will return false and
+                 * LastError() will contain the errorCode.
+                 */
+                Uri(aws_uri_builder_options &builderOptions, Allocator *allocator = DefaultAllocator()) noexcept;
+                Uri(const Uri &) = delete;
+                Uri &operator=(const Uri &) = delete;
+                Uri(Uri &&uri) noexcept;
+                Uri &operator=(Uri &&) noexcept;
+
+                operator bool() const noexcept { return m_isInit; }
```
marcomagdy: make it explicit

samples/mqtt_pub_sub/main.cpp
```diff
@@ -327,15 +306,29 @@ int main(int argc, char *argv[])
...
-            connection->Publish(topic.c_str(), AWS_MQTT_QOS_AT_LEAST_ONCE, false, payload, onOpComplete);
+            ByteBuf payload = ByteBufNewCopy(DefaultAllocator(), (const uint8_t *)input.data(), input.length());
+            ByteBuf *payloadPtr = &payload;
+
+            auto onPublishComplete = [payloadPtr](Mqtt::MqttConnection &, uint16_t packetId, int errorCode) {
+                aws_byte_buf_clean_up(payloadPtr);
+
+                if (packetId)
+                {
+                    fprintf(stdout, "Operation on packetId %d Succeeded\n", (int)packetId);
```
marcomagdy: is this cast necessary?

source/http/HttpConnection.cpp
```diff
@@ -0,0 +1,363 @@
...
+                    toSeat->m_onStreamOutgoingBody = requestOptions.onStreamOutgoingBody;
+
+                    wrapper->allocator = m_allocator;
+                    options.user_data = wrapper;
+                    toSeat->m_stream = aws_http_stream_new_client_request(&options);
+
+                    if (!toSeat->m_stream)
+                    {
+                        wrapper->stream.reset();
+                        wrapper->stream = nullptr;
```
marcomagdy: reset and assigning nullptr are the same. No point in having both.

source/io/Uri.cpp
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    m_isInit = true;
+                }
+                else
+                {
+                    m_lastError = aws_last_error();
+                }
+            }
+
+            Uri::Uri(Uri &&uri) noexcept : m_lastError(AWS_ERROR_SUCCESS), m_isInit(false)
```
marcomagdy: should be `m_isInit(uri.m_IsInit)`

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+                ~HttpStream();
+                HttpStream(const HttpStream &) = delete;
+                HttpStream(HttpStream &&) = delete;
+                HttpStream &operator=(const HttpStream &) = delete;
+                HttpStream &operator=(HttpStream &&) = delete;
+
+                /**
+                 * Get the underlying connection for the stream.
+                 */
+                const std::shared_ptr<HttpConnection> &GetConnection() const noexcept;
```
marcomagdy: Is it invariant that an HttpStream always have a connection? if yes, then return by reference instead of shared_ptr

include/aws/crt/http/HttpConnection.h
graebm: You've got one callback for N possible connections. Each one might need its own unique user_data

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+            /**
+             * Called as headers are received from the peer. `headersArray` will contain the header value
+             * read from the wire. The number of entries in `headersArray` are specified in `headersCount`.
+             *
+             * Keep in mind that this function will likely be called multiple times until all headers are received.
+             *
+             * On HttpStream, this function must be set.
+             */
+            using OnIncomingHeaders = std::function<void(
+                const std::shared_ptr<HttpStream> &stream,
```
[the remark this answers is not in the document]
JonathanHenson: yes

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+                std::shared_ptr<HttpStream> NewStream(const HttpRequestOptions &requestOptions) noexcept;
+
+                /**
+                 * Initiate a shutdown of the connection. Sometimes, connections are persistent and you want
+                 * to close them before shutting down your application or whatever is consuming this interface.
+                 *
+                 * Assuming `OnConnectionShutdown` has not already been invoked, it will be invoked as a result of this
+                 * call. It is safe to release your reference to this object after calling this function.
+                 */
+                bool Close() noexcept;
```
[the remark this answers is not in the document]
JonathanHenson: not at the moment

include/aws/crt/http/HttpConnection.h
```diff
@@ -0,0 +1,356 @@
...
+                    int error_code,
+                    void *user_data) noexcept;
+
+                friend class HttpConnection;
+            };
+
+            /**
+             * Represents a connection between an Http Client and Server.
+             */
+            class AWS_CRT_CPP_API HttpConnection final : public std::enable_shared_from_this<HttpConnection>
```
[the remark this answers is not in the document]
JonathanHenson: resolved for now, It's now HttpClientConnection and HttpClientStream inherits from HttpStream

source/http/HttpConnection.cpp
```diff
@@ -21,158 +21,120 @@ namespace Aws
...
-                int retError = errorCode;
+                auto *callbackData = static_cast<ConnectionCallbackData *>(user_data);
                 if (!errorCode)
                 {
-                    Allocator *allocator = connectionWrapper->client->m_allocator;
+                    Allocator *allocator = callbackData->allocator;
 
+                    /* Why aren't we using New...? Because the constructor is private and the C++ type system
+                     * isn't the boss of me. */
                     auto *toSeat =
```
justinboswell: This should use New()

source/http/HttpConnection.cpp
```diff
@@ -206,56 +169,51 @@ namespace Aws
                 options.on_complete = HttpStream::s_onStreamComplete;
                 options.client_connection = m_connection;
 
-                /* Do the same ref counting trick we did with HttpConnection. We need to maintain a reference
+                /* Do the same ref counting trick we did with HttpClientConnection. We need to maintain a reference
                  * internally (regardless of what the user does), until the Stream shuts down. */
-                auto *toSeat = reinterpret_cast<HttpStream *>(aws_mem_acquire(m_allocator, sizeof(HttpStream)));
+                auto *toSeat = static_cast<HttpClientStream *>(aws_mem_acquire(m_allocator, sizeof(HttpStream)));
```
justinboswell: This should use New()

source/http/HttpConnection.cpp
JonathanHenson: Can’t, Private constructor

source/http/HttpConnection.cpp
JonathanHenson: Private constructor

include/aws/crt/io/Uri.h
```diff
@@ -0,0 +1,102 @@
...
+
+                /** Returns the Path and Query portion of the URI. In the case of Http, this likely the value for the
+                 * URI parameter.
+                 */
+                ByteCursor GetPathAndQuery() const noexcept;
+
+                /**
+                 * The full URI as it was passed to or parsed from the constructors.
+                 */
+                const ByteBuf &GetFullUri() const noexcept;
```
graebm: Just return a ByteCursor, like every other Get() function is returning

include/aws/crt/io/Uri.h
```diff
@@ -0,0 +1,102 @@
...
+                 * will contain the errorCode.
+                 */
+                Uri(const ByteCursor &cursor, Allocator *allocator = DefaultAllocator()) noexcept;
+                /**
+                 * builds a URI from `builderOptions`. Upon failure the bool() operator will return false and
+                 * LastError() will contain the errorCode.
+                 */
+                Uri(aws_uri_builder_options &builderOptions, Allocator *allocator = DefaultAllocator()) noexcept;
+                Uri(const Uri &) = delete;
+                Uri &operator=(const Uri &) = delete;
```
graebm: why delete the copy-constructor?

couldn't it just do `Uri(rhs.GetFullUri(), rhs.m_allocator)

include/aws/crt/io/Uri.h
```diff
@@ -0,0 +1,102 @@
...
+                Uri(Uri &&uri) noexcept;
+                Uri &operator=(Uri &&) noexcept;
```
graebm: If you deleted the `operator=(&&)`, this would be an immutable class

not sure which is better, but whatever you choose, do the same with vanilla `operator=(&)`

PR #17, opened 2019-04-02
Fixes to make pubsub sample compatible with greengrasscore.

JonathanHenson:


samples/mqtt_pub_sub/main.cpp
```diff
@@ -74,6 +75,7 @@ int main(int argc, char *argv[])
     String keyPath;
     String caFile;
     String topic;
+    String clientId("client_12345");
```
graebm: We've been using "samples-client-id" as the default in the language SDKs. Let's us use a specific string in the policy.

PR #18, opened 2019-04-04
Connection manager

JonathanHenson:
HttpClientConnection manager, for managing lifetimes and pooling of connections.

This PR should fail until a bug fix that one of these tests found is fixed in aws-c-io. I'll push the tag update once the aws-c-io bug is fixed.


source/http/HttpConnection.cpp
```diff
@@ -81,6 +81,7 @@ namespace Aws
                 (void)connection;
                 /* now that we're shutting down, we can release the internal ref count. */
                 auto *callbackData = static_cast<ConnectionCallbackData *>(user_data);
+                callbackData->connection->m_good = false;
```
graebm: We could add `aws_http_connection_good()` or something like that, which would get you *slightly* more accurate answers. (at the cost of a lock/check/unlock)

Since m_good isn't set false until connection_shutdown(), there's a chance of someone's stream failing, which triggers the connection to close internally, which calls stream_complete() for all the pending streams before connection_shutdown() is invoked. So the user's stream will fail, and they might give the connection back to the ConnectionMgr, but Connection->m_good is still true so ConnectionMgr vends the connection to another user, who can't use it because it's mid-shutdown.

source/http/HttpConnection.cpp
```diff
@@ -130,7 +131,7 @@ namespace Aws
             }
 
             HttpClientConnection::HttpClientConnection(aws_http_connection *connection, Allocator *allocator) noexcept
-                : m_connection(connection), m_allocator(allocator), m_lastError(AWS_ERROR_SUCCESS)
+                : m_connection(connection), m_allocator(allocator), m_lastError(AWS_ERROR_SUCCESS), m_good(true)
```
justinboswell: For clarity, perhaps call ```m_good``` ```m_open``` or something more descriptive?

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+                    {
+                        m_lastError = aws_last_error();
+                        m_good = false;
+                    }
+                }
+            }
+
+            HttpClientConnectionManager::~HttpClientConnectionManager()
+            {
+                Vector<std::shared_ptr<HttpClientConnection>> connectionsCopy = m_connections;
```
justinboswell: Shouldn't you hold ```m_connectionsLock``` while copying this?

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+                    /* wait for the last one to shutdown so the memory will be cleaned up. */
+                    std::unique_lock<std::mutex> m_semaphoreLock(m_connectionsLock);
+                    m_shutdownSemaphore.wait(m_semaphoreLock, [this] { return m_connections.empty(); });
+                }
+
+                /*
+                 * This would be really screwy if this ever happened. I don't even know what error to report, but in
+                 * case someone is waiting for a connection, AT LEAST let them know so they can not deadlock.
+                 */
+                for (auto &pendingConnectionRequest : m_pendingConnectionRequests)
```
justinboswell: Should ```m_connectionsLock``` be held while you do this?

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+                {
+                    pendingConnectionRequest(nullptr, AWS_OP_ERR);
+                }
+            }
+
+            /* Asssumption: Whoever calls this already holds the lock. */
+            bool HttpClientConnectionManager::s_createConnection() noexcept
+            {
+                if (m_connections.size() + m_outstandingVendedConnections + m_pendingConnections < m_max_size)
```
justinboswell: ```m_maxSize```

Your C is showing ;)

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+                {
+                    HttpClientConnectionOptions connectionOptions;
+                    connectionOptions.socketOptions = &m_socketOptions;
+                    connectionOptions.port = m_port;
+                    connectionOptions.hostName = ByteCursorFromCString(m_hostName.c_str());
+                    connectionOptions.bootstrap = m_bootstrap;
+                    connectionOptions.initialWindowSize = m_initialWindowSize;
+                    connectionOptions.allocator = m_allocator;
+                    connectionOptions.onConnectionSetup =
+                        [this](const std::shared_ptr<HttpClientConnection> &connection, int errorCode) {
```
justinboswell: You aren't using this, and the callback is static, so you don't need the capture

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+                            s_onConnectionSetup(connection, errorCode);
+                        };
+                    connectionOptions.onConnectionShutdown = [this](HttpClientConnection &connection, int errorCode) {
```
justinboswell: Same here, [this] isn't needed

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+                /* make sure all connections we know about are closed. */
+                for (auto &connection : connectionsCopy)
+                {
+                    connection->Close();
+                }
+
+                {
+                    /* wait for the last one to shutdown so the memory will be cleaned up. */
+                    std::unique_lock<std::mutex> m_semaphoreLock(m_connectionsLock);
+                    m_shutdownSemaphore.wait(m_semaphoreLock, [this] { return m_connections.empty(); });
```
graebm: This can deadlock if the ConnectionMgr is released from an event-loop thread. And we want users to set their apps up as event-driven, so that's pretty likely isn't it?

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+                std::shared_ptr<HttpClientConnection> connection,
+                bool isRelease) noexcept
+            {
+                /** Locking logic:
+                 *  We can't hold the lock during the callback since a user may call back into the connection manager
+                 *  from the callbacks. We don't use a scoped lock because we need finer grain control.
+                 *  All unlocks, should either be followed immediately by a callback and then a return, or
+                 *  a return only. If you see anything else, it's a bug.
+                 */
+                m_connectionsLock.lock();
```
justinboswell: Another way you can handle this locking situation (to improve clarity and reduce peril for maintenance) is to use the Synchronized trick (you can find examples in Abseil and Folly);

Basically, associate the mutex with a structure that you want to modify under the mutex's protection.

```cpp
template <typename T, typename Mutex>
class Synchronized
{
public:
    void Modify(std::function<void(T&)> modify) {
        std::lock_guard<Mutex> lock(&m_mutex);
        modify(m_t);
    }
private:
    T m_t;
    Mutex m_mutex;
}
```
And then stick the protected data in a structure like so:
```cpp
struct {
    m_connections;
    m_pendingConnectionRequests;
    m_pendingConnections;
    m_outstandingVendedConnections
}
```

And then pass this logic in as lambdas to Modify

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,294 @@
...
+                    connectionOptions.hostName = ByteCursorFromCString(m_hostName.c_str());
+                    connectionOptions.bootstrap = m_bootstrap;
+                    connectionOptions.initialWindowSize = m_initialWindowSize;
+                    connectionOptions.allocator = m_allocator;
+                    connectionOptions.onConnectionSetup =
+                        [this](const std::shared_ptr<HttpClientConnection> &connection, int errorCode) {
+                            s_onConnectionSetup(connection, errorCode);
+                        };
+                    connectionOptions.onConnectionShutdown = [this](HttpClientConnection &connection, int errorCode) {
+                        s_onConnectionShutdown(connection, errorCode);
```
graebm: here's a possible solution for the manager-must-outlive-its-connections problem:
```
std::weak_ptr<ConnectionManager> weakThis = weak_from_this();
connectionOptions.onConnectionShutdown = [weakThis](HttpClientConnection &connection, int errorCode) {
    if (auto sharedThis = weakThis.lock()) {
        sharedThis.s_onConnectionShutdown(connection, errorCode);
    }
}
```

The Manager would have to be shared_from_this, but I think this is pretty graceful.
This would remove the need for the semaphore stuff the Manager's destructor

include/aws/crt/Types.h
```diff
@@ -62,6 +63,7 @@ namespace Aws
         template <typename K, typename V>
         using MultiMap = std::multimap<K, V, std::less<K>, StlAllocator<std::pair<const K, V>>>;
         template <typename T> using Vector = std::vector<T, StlAllocator<T>>;
+        template <typename T> using List = std::list<T, StlAllocator<T>>;
```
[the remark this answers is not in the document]
JonathanHenson: clang-format

source/http/HttpConnectionManager.cpp
JonathanHenson: wasn't needed anyways.

source/http/HttpConnection.cpp
JonathanHenson: we'll come back to this added a TODO

tests/HttpClientTest.cpp
```diff
@@ -108,105 +100,97 @@ static int s_TestHttpDownloadNoBackPressure(struct aws_allocator *allocator, voi
...
-    connection->NewClientStream(requestOptions);
-    semaphore.wait(semaphoreULock, [&]() { return streamCompleted; });
-    ASSERT_INT_EQUALS(200, responseCode);
-
-    connection->Close();
-    semaphore.wait(semaphoreULock, [&]() { return connectionShutdown; });
-
-    downloadedFile.flush();
-    downloadedFile.close();
+    {
```
graebm: don't modify HttpClientTest.cpp to use the ConnectionManager

JonathanHenson: Will revert

include/aws/crt/io/HostResolver.h
```diff
@@ -0,0 +1,92 @@
...
+            class EventLoopGroup;
+            class HostResolver;
+
+            /**
+             * Invoked upon resolution of an address. You do not own the memory pointed to in addresses, if you persist
+             * the data, copy it first. If errorCode is AWS_ERROR_SUCCESS, the operation succeeded. Otherwise, the
+             * operation failed.
+             */
+            using OnHostResolved =
+                std::function<void(HostResolver &resolver, const Vector<aws_host_address> &addresses, int errorCode)>;
```
justinboswell: Should we type alias aws_host_address?

include/aws/crt/io/HostResolver.h
```diff
@@ -0,0 +1,92 @@
...
+
+            class HostResolver
+            {
+              public:
+                virtual ~HostResolver();
```
justinboswell: = default since the impl is empty?

include/aws/crt/io/HostResolver.h
```diff
@@ -0,0 +1,92 @@
...
+                 */
+                bool ResolveHost(const String &host, const OnHostResolved &onResolved) noexcept override;
+                aws_host_resolver *GetUnderlyingHandle() noexcept override { return &m_resolver; }
+                aws_host_resolution_config *GetConfig() noexcept override { return &m_config; }
+
+              private:
+                aws_host_resolver m_resolver;
+                aws_host_resolution_config m_config;
+                Allocator *m_allocator;
+                bool m_good;
```
justinboswell: [nitpick] m_running? m_valid? m_initialized?

include/aws/crt/io/HostResolver.h
JonathanHenson: IIRC, that breaks linkers.

include/aws/crt/io/HostResolver.h
JonathanHenson: m_initialized is fine I suppose.

include/aws/crt/io/HostResolver.h
JonathanHenson: Sure.

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,349 @@
...
+                        userCallback = m_pendingConnectionRequests.front();
+                        m_pendingConnectionRequests.pop_front();
+                        ++m_outstandingVendedConnections;
+                        callbackError = AWS_ERROR_SUCCESS;
+                        vendedConnection = std::move(connection);
+                    }
+                    else if (poolOrVend)
+                    {
+                        /* Case 2.2 */
+                        m_connections.push_back(std::move(connection));
```
graebm: trivial: emplace_back

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,349 @@
...
+                    {
+                        userCallbackNewConnectionFailures.push_back(std::pair<OnClientConnectionAvailable, int>(
+                            m_pendingConnectionRequests.front(), errorCode));
+                        m_pendingConnectionRequests.pop_front();
+                    }
+
+                    /* if we had more pending connection requests than the pool size, we need to replace this
+                       connection. Just loop until we've matched either the connectionRequest to the max amount of
+                       connections in flight, or when run out of pending connection requests. */
+                    while (m_pendingConnectionRequests.size() > m_pendingConnections &&
```
graebm: This case is  **very** similar to the "case 1.2" in `poolOrVendConnection()`, but we handle it differently here in `onConnectionSetup()`

In `poolOrVendConnection()` we try **once** to create a new connection to replace the broken one. Here in `onConnectionSetup()` we try **repeatedly** to create a new connection to replace the broken one.

I have a hunch that trying just once is the better approach, if for no other reason than it's simpler. But maybe there's some complication I'm not thinking of?

source/http/HttpConnectionManager.cpp
```diff
@@ -0,0 +1,349 @@
...
+                           m_connections.size() + m_pendingConnections + m_outstandingVendedConnections < m_maxSize)
+                    {
+                        if (!createConnection())
+                        {
+                            userCallbackNewConnectionFailures.push_back(std::pair<OnClientConnectionAvailable, int>(
+                                m_pendingConnectionRequests.front(), aws_last_error()));
```
graebm: trivial: could use emplace_back(), and could std::move(m_pendingConnectionRequests.front())

PR #31, opened 2019-06-14
Updated dependencies to latest versions.

JonathanHenson:

### DIFF
--- a/aws-common-runtime/CMakeLists.txt
+++ b/aws-common-runtime/CMakeLists.txt
@@ -17,7 +17,7 @@ set(AWS_DEPS_DOWNLOAD_DIR "${AWS_DEPS_BUILD_DIR}/downloads" CACHE PATH "Dependen
 
 message("install dir ${AWS_DEPS_INSTALL_DIR}")
 set(AWS_C_COMMON_URL "https://github.com/awslabs/aws-c-common.git")
-set(AWS_C_COMMON_SHA "v0.3.11")
+set(AWS_C_COMMON_SHA "v0.3.13")
 include(BuildAwsCCommon)
 
 if (UNIX AND NOT APPLE)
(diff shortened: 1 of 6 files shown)


aws-common-runtime/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ set(AWS_DEPS_DOWNLOAD_DIR "${AWS_DEPS_BUILD_DIR}/downloads" CACHE PATH "Dependen
 
 message("install dir ${AWS_DEPS_INSTALL_DIR}")
 set(AWS_C_COMMON_URL "https://github.com/awslabs/aws-c-common.git")
-set(AWS_C_COMMON_SHA "v0.3.11")
+set(AWS_C_COMMON_SHA "v0.3.13")
```
justinboswell: There's now a v0.3.14, which includes fixes for callstack decoding tests on linux. Should be a trivial update.

source/http/HttpConnection.cpp
```diff
@@ -39,6 +39,24 @@ namespace Aws
                 OnConnectionShutdown onConnectionShutdown;
             };
 
+            class UnManagedConnection final : public HttpClientConnection
```
justinboswell: Unmanaged, perhaps?

tests/HttpClientConnectionManagerTest.cpp
```diff
@@ -113,7 +113,7 @@ static int s_TestHttpClientConnectionManagerResourceSafety(struct aws_allocator
     connections.clear();
     for (auto &connection : connectionsCpy)
```
justinboswell: clearing connectionsCpy would have the same effect, wouldn't it?

JonathanHenson: oh.... yeah true.

PR #37, opened 2019-06-27
Allocator rework

bretambrose:
Add stateful reference to underlying c-common allocator.

An internal #define to make the default constructor invisible verifies that crt logic always manually selects an aws-c-common allocator.

There are some potentially questionable decision made here.

Also, certain construction paths (in particular, Vector<String> (count)) required a code change because the internal implementation was not cloning the Vector's allocator when in-place building the reserved Strings.  This makes me nervous.

This will require a new release

### DIFF
--- a/include/aws/crt/JsonObject.h
+++ b/include/aws/crt/JsonObject.h
@@ -33,12 +33,12 @@ namespace Aws
             /**
              * Constructs empty JSON DOM.
              */
-            JsonObject();
+            JsonObject(Allocator *a = DefaultAllocator());
 
             /**
              * Constructs a JSON DOM by parsing the input string.
              */
-            JsonObject(const String &value);
+            JsonObject(const String &value, Allocator *a = DefaultAllocator());
 
             /**
              * Performs a deep copy of the JSON DOM parameter.
@@ -200,7 +200,8 @@ namespace Aws
 
           private:
             void Destroy();
-            JsonObject(cJSON *value);
+            JsonObject(cJSON *value, Allocator *a);
+            Allocator *m_allocator;
             cJSON *m_value;
             bool m_wasParseSuccessful;
             String m_errorMessage;
@@ -217,7 +218,7 @@ namespace Aws
         {
           public:
             /* constructors */
-            JsonView();
+            JsonView(Allocator *a = DefaultAllocator());
             JsonView(const JsonObject &val);
             JsonView &operator=(const JsonObject &val);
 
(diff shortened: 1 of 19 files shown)


include/aws/iot/MqttClient.h
```diff
@@ -29,7 +29,7 @@ namespace Aws
         class AWS_CRT_CPP_API MqttClientConnectionConfig final
         {
           public:
-            MqttClientConnectionConfig() noexcept;
+            // MqttClientConnectionConfig() noexcept;
```
justinboswell: Should this just be removed?

include/aws/crt/JsonObject.h
```diff
@@ -33,12 +33,12 @@ namespace Aws
             /**
              * Constructs empty JSON DOM.
              */
-            JsonObject();
+            JsonObject(Allocator *a = DefaultAllocator());
```
justinboswell: Allocators are usually taken by reference, as they cannot be null. Why not do that for these objects?

include/aws/crt/Types.h
```diff
@@ -76,8 +76,8 @@ namespace Aws
         AWS_CRT_CPP_API ByteCursor ByteCursorFromByteBuf(const ByteBuf &) noexcept;
         AWS_CRT_CPP_API ByteCursor ByteCursorFromArray(const uint8_t *array, size_t len) noexcept;
 
-        AWS_CRT_CPP_API Vector<uint8_t> Base64Decode(const String &decode);
-        AWS_CRT_CPP_API String Base64Encode(const Vector<uint8_t> &encode);
+        AWS_CRT_CPP_API Vector<uint8_t> Base64Decode(const String &decode, Allocator *a = DefaultAllocator());
```
justinboswell: Allocator should be a reference.

include/aws/crt/UUID.h
```diff
@@ -25,7 +25,7 @@ namespace Aws
         class UUID final
         {
           public:
-            UUID() noexcept;
+            UUID(Allocator *a = DefaultAllocator()) noexcept;
```
justinboswell: Allocator should be a reference

include/aws/iot/MqttClient.h
bretambrose: Yeah, oops

include/aws/crt/JsonObject.h
bretambrose: Mostly just following the existing pattern established in DefaultHostResolver, HttpConnectionManager, etc...  These are the c versions.

PR #40, opened 2019-07-09
Namespace wrap CJSON

bretambrose:

### DIFF
--- a/include/aws/crt/external/cJSON.h
+++ b/include/aws/crt/external/cJSON.h
@@ -23,6 +23,7 @@
 /** MODIFICATIONS:
  * parse doesn't rely on null terminated strings. (Who thought that was a good idea.... ever?).
  * valueInt was moved up to improve alignment.
+ * Wrap all symbols in the Aws namespace as a short-term collision resolution
  *
  * Modifications licensed under:
  *
--- a/include/aws/crt/external/cJSON.h
+++ b/include/aws/crt/external/cJSON.h
@@ -43,10 +43,7 @@
 #ifndef cJSON__h
 #define cJSON__h
 
-#ifdef __cplusplus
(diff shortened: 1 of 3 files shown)


include/aws/crt/external/cJSON.h
```diff
@@ -152,148 +149,191 @@ then using the CJSON_API_VISIBILITY flag to "export" the same symbols the way CJ
...
-CJSON_PUBLIC(cJSON_bool) cJSON_IsTrue(const cJSON * const item);
-CJSON_PUBLIC(cJSON_bool) cJSON_IsBool(const cJSON * const item);
-CJSON_PUBLIC(cJSON_bool) cJSON_IsNull(const cJSON * const item);
-CJSON_PUBLIC(cJSON_bool) cJSON_IsNumber(const cJSON * const item);
-CJSON_PUBLIC(cJSON_bool) cJSON_IsString(const cJSON * const item);
-CJSON_PUBLIC(cJSON_bool) cJSON_IsArray(const cJSON * const item);
-CJSON_PUBLIC(cJSON_bool) cJSON_IsObject(const cJSON * const item);
-CJSON_PUBLIC(cJSON_bool) cJSON_IsRaw(const cJSON * const item);
+    CJSON_PUBLIC(cJSON_bool) cJSON_IsInvalid(const cJSON *const item);
+
```
justinboswell: Where did all these whitespace changes come from? Is this just from adding the namespace scope, and then the IDE went mad with power? Could we add only the namespace/scope lines?

bretambrose: Not if we want format-check to succeed =/

justinboswell: :(

PR #41, opened 2019-07-10
Integrate aws-c-auth; logging; misc tweaks

bretambrose:
Aws-c-auth binding preliminaries:

* Integrate into build
* Simple logging init
* A few miscellaneous initialization uptdates


source/Api.cpp
```diff
@@ -46,21 +54,53 @@ namespace Aws
...
-        ApiHandle::~ApiHandle() { s_cleanUpApi(); }
+            if (aws_logger_get() == &logger)
+            {
+                aws_logger_set(NULL);
+                aws_logger_clean_up(&logger);
+            }
+        }
 
-        void LoadErrorStrings() noexcept
+        void ApiHandle::InitializeLogging(Aws::Crt::LogLevel level, const char *filename)
```
JonathanHenson: it might be good to accept a file pointer as well so we can dump to stderr or stdout etc....

bretambrose: Added an overload that takes a file pointer and refactored to make nicer.

PR #42, opened 2019-07-10
Credentials and credentials provider bindings

bretambrose:

### DIFF
--- a/source/auth/Credentials.cpp
+++ b/source/auth/Credentials.cpp
@@ -43,22 +43,49 @@ namespace Aws
             {
             }
 
-            Credentials::~Credentials() { aws_credentials_destroy(m_credentials); }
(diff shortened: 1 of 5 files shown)


source/auth/Credentials.cpp
```diff
@@ -0,0 +1,252 @@
...
+            {
+            }
+
+            Credentials::~Credentials()
+            {
+                aws_credentials_destroy(m_credentials);
+                m_credentials = nullptr;
+            }
+
+            ByteCursor Credentials::GetAccessKeyId(void) const noexcept
```
JonathanHenson: do you need (void) in C++?

bretambrose: No.  Removed.

PR #44, opened 2019-07-11
Wrap aws_http_request

bretambrose:


include/aws/crt/http/HttpRequest.h
```diff
@@ -0,0 +1,70 @@
...
+                std::shared_ptr<Aws::Crt::Io::IStream> GetBody() const noexcept;
+                bool SetBody(const std::shared_ptr<Aws::Crt::Io::IStream> &body) noexcept;
+
+                size_t GetHeaderCount() const noexcept;
+                Optional<HttpHeader> GetHeader(size_t index) const noexcept;
+                bool SetHeader(size_t index, const HttpHeader &header) noexcept;
+                bool AddHeader(const HttpHeader &header) noexcept;
+                bool EraseHeader(size_t index) noexcept;
+
+                operator bool() const noexcept { return m_request != nullptr; }
```
graebm: There's no m_lastError on this class, like other classes in this SDK

but I agree with you that probably none of these classes should have a local m_lastError, users should just query the global-thread-local one

source/http/HttpRequest.cpp
```diff
@@ -0,0 +1,117 @@
...
+                    stream = Aws::Crt::Io::AwsInputStreamNewCpp(body, m_allocator);
+                    if (stream == nullptr)
+                    {
+                        return false;
+                    }
+                }
+
+                aws_http_request_set_body_stream(m_request, stream);
+
+                m_bodyStream = (stream) ? body : nullptr;
```
graebm: I don't think the aws_input_stream ever gets deleted
You need to delete it upon destruction, and you need to delete it if someone ever calls SetBody() a 2nd time

PR #45, opened 2019-07-13
Percolate TLS setup errors

graebm:
Previously, errors from building the TLS context options didn't percolate all the way up to where the user could see them.

This was cause leading to cases where the wrong kind of TLS connection was being established without realizing it: https://github.com/awslabs/aws-iot-device-sdk-cpp-v2/issues/50


source/iot/MqttClient.cpp
```diff
@@ -148,25 +186,42 @@ namespace Aws
...
-                return newConnection;
+                m_lastError = newConnection->LastError();
+                return nullptr;
             }
 
-            return nullptr;
+            return newConnection;
         }
     } // namespace Iot
 } // namespace Aws
```
JonathanHenson: newline

include/aws/iot/MqttClient.h
```diff
@@ -29,20 +29,24 @@ namespace Aws
         class AWS_CRT_CPP_API MqttClientConnectionConfig final
         {
           public:
-            MqttClientConnectionConfig() noexcept;
+            static MqttClientConnectionConfig CreateInvalid(int lastError) noexcept;
```
JonathanHenson: why is this function public?

graebm: MqttClientConnectionConfigBuilder creates the MqttClientConnectionConfig, but isn't friends, so this is public

PR #51, opened 2019-08-09
Byte buf proposal

bretambrose:
Attempt 2


include/aws/crt/Outcome.h
```diff
@@ -0,0 +1,157 @@
...
+
+            const R &GetResult() const { return m_result; }
+
+            R &GetResult() { return m_result; }
+
+            /**
+             * casts the underlying result to an r-value so that caller can take ownership of underlying resources.
+             * this is necessary when streams are involved.
+             */
+            R &&AcquireResult() { return std::move(m_result); }
```
graebm: I think this helper function is more confusing than just letting users call `std::move(outcome.GetResult())`.

The standard library doesn't offer any helper functions like this. If you want to acquire the contents of an optional<T> you just do `std::move(optional.value())` or `std::move(*optional)`

include/aws/crt/Outcome.h
```diff
@@ -0,0 +1,157 @@
...
+
+            const E &GetError() const { return m_error; }
+
+            operator bool() const { return m_success; }
+
+          private:
+            R m_result;
+            E m_error;
```
graebm: problems for another day: Both R and E types need to be default-constructible for this to work. Like, how will this work with HttpConnection whose default-constructor will be private?

include/aws/crt/Outcome.h
```diff
@@ -0,0 +1,157 @@
...
+    {
+        template <typename R, typename E> // Result, Error
+        class Outcome
+        {
+          public:
+            Outcome() : m_success(false) {}
+
+            Outcome(const R &r) : m_result(r), m_success(true) {}
+
+            Outcome(const E &e) : m_error(e), m_success(false) {}
```
graebm: Problems for another day: Works until R and E might be confused for one another by the compiler.
Ex: `AwsCrtResult<int> HttpClientStream::GetStatusCode()`

graebm (commented):
We don't need to address all these edge-cases. I'm just being a know-it-all who also wrote an Outcome class once.

include/aws/crt/Outcome.h
bretambrose: True, but in that case I would assume you'd be getting a pointer back, not an instance.

include/aws/crt/Outcome.h
bretambrose: Updated

include/aws/crt/Outcome.h
bretambrose: True, but I wish we had a specific error type other than int as well, which would make things much less likely.

bretambrose:
On second thought, let's not.

PR #52, opened 2019-08-12
Implementation + tests only

bretambrose:
rest of project does not build without usage updates

### DIFF
--- a/include/aws/crt/ByteBuf.h
+++ b/include/aws/crt/ByteBuf.h
@@ -31,15 +31,27 @@ namespace Aws
             explicit ByteCursor(const char *str) noexcept;
             explicit ByteCursor(const String &str) noexcept;
             explicit ByteCursor(aws_byte_cursor cursor) noexcept;
-            explicit ByteCursor(aws_byte_cursor *cursor) noexcept;
             explicit ByteCursor(const aws_byte_buf &buffer) noexcept;
-            explicit ByteCursor(const aws_byte_buf *buffer) noexcept;
             ByteCursor(const uint8_t *array, size_t len) noexcept;
 
             ByteCursor &operator=(const ByteCursor &cursor) noexcept;
 
-            aws_byte_cursor *Get() noexcept { return m_cursorPtr; }
-            const aws_byte_cursor *Get() const noexcept { return m_cursorPtr; }
+            /**
+             * This creates an instance that's a reference to the passed in
+             * cursor rather than copying it.  Side-effects to this object
+             * will affect the original cursor.
+             */
+            static ByteCursor Wrap(aws_byte_cursor *cursor) noexcept;
+
+            // APIs
+            void Advance(size_t len) noexcept;
+
+            // Accessors
+            aws_byte_cursor *GetImpl() noexcept { return m_cursorPtr; }
+            const aws_byte_cursor *GetImpl() const noexcept { return m_cursorPtr; }
+
+            const uint8_t *GetPtr() const noexcept { return m_cursorPtr->ptr; }
+            size_t GetLen() const noexcept { return m_cursorPtr->len; }
 
           private:
             aws_byte_cursor m_cursor;
@@ -52,26 +64,34 @@ namespace Aws
             // Initialization that cannot fail
             ByteBuf() noexcept;
             ByteBuf(ByteBuf &&rhs) noexcept;
-            explicit ByteBuf(aws_byte_buf *buffer) noexcept;
-            explicit ByteBuf(const char *str) noexcept;
             ByteBuf(const uint8_t *array, size_t capacity, size_t len) noexcept;
 
             ByteBuf &operator=(ByteBuf &&buffer) noexcept;
 
             ~ByteBuf();
 
+            /**
+             * This creates an instance that is a reference to the passed in
+             * buffer rather than copying it.  Side-effects to this object
+             * will affect the original c buffer.
+             */
+            static ByteBuf Wrap(aws_byte_buf *buffer) noexcept;
+
             // Initialization that can fail
             static AwsCrtResult<ByteBuf> Init(const ByteBuf &buffer) noexcept;
             static AwsCrtResult<ByteBuf> Init(Allocator *alloc, size_t capacity) noexcept;
-            static AwsCrtResult<ByteBuf> InitFromArray(Allocator *alloc, const uint8_t *array, size_t len) noexcept;
 
             // All of non-init APIs go here
             AwsCrtResult<void> Append(ByteCursor cursor) noexcept;
             AwsCrtResult<void> AppendDynamic(ByteCursor cursor) noexcept;
             // etc...
 
-            aws_byte_buf *Get() noexcept { return m_bufferPtr; }
-            const aws_byte_buf *Get() const noexcept { return m_bufferPtr; }
+            aws_byte_buf *GetImpl() noexcept { return m_bufferPtr; }
+            const aws_byte_buf *GetImpl() const noexcept { return m_bufferPtr; }
+
+            const uint8_t *GetBuffer() const noexcept { return m_bufferPtr->buffer; }
+            size_t GetLen() const noexcept { return m_bufferPtr->len; }
+            size_t GetCapacity() const noexcept { return m_bufferPtr->capacity; }
 
             ByteCursor GetCursor() const noexcept;
 
@@ -84,5 +104,6 @@ namespace Aws
             aws_byte_buf m_buffer;
             aws_byte_buf *m_bufferPtr;
         };
+
     } // namespace Crt
 } // namespace Aws
--- a/include/aws/crt/ByteBuf.h
+++ b/include/aws/crt/ByteBuf.h
@@ -1,0 +1,88 @@
+#pragma once
+/*
+ * Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/crt/Outcome.h>
+#include <aws/crt/Types.h>
+
+#include <aws/common/byte_buf.h>
+
+namespace Aws
+{
+    namespace Crt
+    {
+        class ByteCursor
+        {
+          public:
+            ByteCursor() noexcept;
+            ByteCursor(const ByteCursor &cursor) noexcept;
+            explicit ByteCursor(const char *str) noexcept;
+            explicit ByteCursor(const String &str) noexcept;
+            explicit ByteCursor(aws_byte_cursor cursor) noexcept;
+            explicit ByteCursor(aws_byte_cursor *cursor) noexcept;
+            explicit ByteCursor(const aws_byte_buf &buffer) noexcept;
+            explicit ByteCursor(const aws_byte_buf *buffer) noexcept;
+            ByteCursor(const uint8_t *array, size_t len) noexcept;
+
+            ByteCursor &operator=(const ByteCursor &cursor) noexcept;
+
+            aws_byte_cursor *Get() noexcept { return m_cursorPtr; }
+            const aws_byte_cursor *Get() const noexcept { return m_cursorPtr; }
+
+          private:
+            aws_byte_cursor m_cursor;
+            aws_byte_cursor *m_cursorPtr;
+        };
+
+        class ByteBuf
+        {
+          public:
+            // Initialization that cannot fail
+            ByteBuf() noexcept;
+            ByteBuf(ByteBuf &&rhs) noexcept;
+            explicit ByteBuf(aws_byte_buf *buffer) noexcept;
+            explicit ByteBuf(const char *str) noexcept;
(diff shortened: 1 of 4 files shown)


include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,88 @@
...
+
+            // Initialization that can fail
+            static AwsCrtResult<ByteBuf> Init(const ByteBuf &buffer) noexcept;
+            static AwsCrtResult<ByteBuf> Init(Allocator *alloc, size_t capacity) noexcept;
+            static AwsCrtResult<ByteBuf> InitFromArray(Allocator *alloc, const uint8_t *array, size_t len) noexcept;
+
+            // All of non-init APIs go here
+            AwsCrtResult<void> Append(ByteCursor cursor) noexcept;
+            AwsCrtResult<void> AppendDynamic(ByteCursor cursor) noexcept;
+            // etc...
```
justinboswell: What about the rest?

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,88 @@
...
+            ByteBuf(const uint8_t *array, size_t capacity, size_t len) noexcept;
+
+            ByteBuf &operator=(ByteBuf &&buffer) noexcept;
+
+            ~ByteBuf();
+
+            // Initialization that can fail
+            static AwsCrtResult<ByteBuf> Init(const ByteBuf &buffer) noexcept;
+            static AwsCrtResult<ByteBuf> Init(Allocator *alloc, size_t capacity) noexcept;
+            static AwsCrtResult<ByteBuf> InitFromArray(Allocator *alloc, const uint8_t *array, size_t len) noexcept;
```
graebm: Likewise, I'd argue that InitFromArray() shouldn't exist. Buffers are for writing.
If there was a need to init a buffer that starts by copying in a bunch of contents, that function should probably accept separate `len` and `capacity` args.

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,88 @@
...
+
+            // All of non-init APIs go here
+            AwsCrtResult<void> Append(ByteCursor cursor) noexcept;
+            AwsCrtResult<void> AppendDynamic(ByteCursor cursor) noexcept;
+            // etc...
+
+            aws_byte_buf *Get() noexcept { return m_bufferPtr; }
+            const aws_byte_buf *Get() const noexcept { return m_bufferPtr; }
```
graebm: byte cursor has `ptr`, `len`, and `capacity` members that we frequently access in C. We should probably make them accessible in this class as well.

Since we'll have other getters, we probably want the functions that return the underlying c-structs to have a more specific name than "Get()". Impl(), GetImpl(), CStruct(), GetStruct(), Repr(), AwsC()?

Unless we're leaving the "Get" out of the other getters and only using that term to refer to the c-struct

source/ByteBuf.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+#include <aws/crt/ByteBuf.h>
+
+namespace Aws
+{
+    namespace Crt
+    {
+        ///////////////////////////////////////////////////////////////////////////////////
+
+        ByteCursor::ByteCursor() noexcept : m_cursorPtr(&m_cursor) { AWS_ZERO_STRUCT(m_cursor); }
```
graebm: trivial: In c++ you can zero out a struct in the member-initializer-list like so
`: m_cursor{0}`

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,88 @@
...
+    {
+        class ByteCursor
+        {
+          public:
+            ByteCursor() noexcept;
+            ByteCursor(const ByteCursor &cursor) noexcept;
+            explicit ByteCursor(const char *str) noexcept;
+            explicit ByteCursor(const String &str) noexcept;
+            explicit ByteCursor(aws_byte_cursor cursor) noexcept;
+            explicit ByteCursor(aws_byte_cursor *cursor) noexcept;
```
graebm: I would be more explicit about what's going on with the by-pointer constructor. It has *radically* different behavior than the rest (modifying the C++ object, also modifies the thing used to construct it). Maybe:

static ByteCursor InitWrapper(aws_byte_cursor *cursorToWrap);

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,88 @@
...
+            explicit ByteCursor(const aws_byte_buf &buffer) noexcept;
+            explicit ByteCursor(const aws_byte_buf *buffer) noexcept;
+            ByteCursor(const uint8_t *array, size_t len) noexcept;
+
+            ByteCursor &operator=(const ByteCursor &cursor) noexcept;
+
+            aws_byte_cursor *Get() noexcept { return m_cursorPtr; }
```
graebm: Probably want to implement aws_byte_cursor_advance()

source/ByteBuf.cpp
```diff
@@ -0,0 +1,230 @@
...
+            if (&rhs != this)
+            {
+                Cleanup();
+                if (rhs.m_bufferPtr == &rhs.m_buffer)
+                {
+                    m_buffer = rhs.m_buffer;
+                    m_bufferPtr = &m_buffer;
+
+                    AWS_ZERO_STRUCT(rhs.m_buffer);
+                    rhs.m_bufferPtr = &rhs.m_buffer;
```
graebm: trivial: this line isn't necessary `rhs.m_bufferPtr = &rhs.m_buffer;`

include/aws/crt/ByteBuf.h
bretambrose: Given the width of the byte buf / cursor API, is it reasonable to just add as needed?

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,88 @@
...
+        };
+
+        class ByteBuf
+        {
+          public:
+            // Initialization that cannot fail
+            ByteBuf() noexcept;
+            ByteBuf(ByteBuf &&rhs) noexcept;
+            explicit ByteBuf(aws_byte_buf *buffer) noexcept;
+            explicit ByteBuf(const char *str) noexcept;
```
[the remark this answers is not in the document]
bretambrose: Updated

include/aws/crt/ByteBuf.h
bretambrose: Removed as well.  Usage replaced with Append

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,109 @@
...
+            /**
+             * This creates an instance that is a reference to the passed in
+             * buffer rather than copying it.  Side-effects to this object
+             * will affect the original c buffer.
+             */
+            static ByteBuf Wrap(aws_byte_buf *buffer) noexcept;
+
+            // Initialization that can fail
+            static AwsCrtResult<ByteBuf> Init(const ByteBuf &buffer) noexcept;
+            static AwsCrtResult<ByteBuf> Init(Allocator *alloc, size_t capacity) noexcept;
```
marcomagdy: I will say that having an init function versus a state variable has one single advantage; and that is it can be marked with `[[nodiscard]]` in the future. That being said, I still so much dislike the ctor + init pattern.

So, what are your thoughts about having an (optional?) error_callback function parameter to the ctors overloads that can fail?

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,109 @@
...
+
+            // All of non-init APIs go here
+            AwsCrtResult<void> Append(ByteCursor cursor) noexcept;
+            AwsCrtResult<void> AppendDynamic(ByteCursor cursor) noexcept;
+            // etc...
+
+            aws_byte_buf *GetImpl() noexcept { return m_bufferPtr; }
+            const aws_byte_buf *GetImpl() const noexcept { return m_bufferPtr; }
```
marcomagdy: would you consider an **explicit** type conversion operator here instead? (note the emphasis on explicit 😄)

source/ByteBuf.cpp
marcomagdy: nitpick: `m_cursor{0}` initializes the **first** member to 0 and default initialize the rest of the members. However, in this case they're the same.
But `m_cursor{42}` would _not_ initialize all the members to 42!!!
Instead, it initializes the first member to 42 and default initialize the rest of the members.

So, `m_cursor{}` is just as good to default initialize the struct. effectively zeroing it out.

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,109 @@
...
+            ByteBuf &operator=(ByteBuf &&buffer) noexcept;
+
+            ~ByteBuf();
+
+            /**
+             * This creates an instance that is a reference to the passed in
+             * buffer rather than copying it.  Side-effects to this object
+             * will affect the original c buffer.
+             */
+            static ByteBuf Wrap(aws_byte_buf *buffer) noexcept;
```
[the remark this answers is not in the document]
bretambrose: Passing clib-created byte bufs up to the SDK.  For example, a callback where the user is supposed to write outgoing data to an internal byte buf.

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,109 @@
...
+#include <aws/crt/Outcome.h>
+#include <aws/crt/Types.h>
+
+#include <aws/common/byte_buf.h>
+
+namespace Aws
+{
+    namespace Crt
+    {
+        class ByteCursor
```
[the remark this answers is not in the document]
bretambrose: There's definitely some tension between idiomatic and thin-wrapper here.  Ultimately, we're not just servicing C++ end-users, we're also servicing those people (mostly ourselves, to be honest) who are writing the glue between the layers.  We can't avoid writing a byte cursor wrapper (see the comment about why we need non-owning bytebuf).

Adding iterators probably makes sense in a future step, and that may be the right interface to expose up to the SDK, but I'm reluctant to make that assumption until we have a better feel for usage patterns beyond the current set of callbacks and thin wrappers.

include/aws/crt/Outcome.h
```diff
@@ -0,0 +1,151 @@
...
+            R &GetResult() { return m_result; }
+
+            const E &GetError() const { return m_error; }
+
+            operator bool() const { return m_success; }
+
+          private:
+            R m_result;
+            E m_error;
+            bool m_success;
```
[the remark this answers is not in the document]
bretambrose: Seems reasonable

include/aws/crt/ByteBuf.h
bretambrose: I don't see how failing assignment could be handled with that approach.  Having a callback that gets invoked in the middle of a construction failure (that still needs to be reacted to afterwards when your object is garbage) seems counter-intuitive.

I'm hoping the same error-handling pattern can be used not just for construction, but for all failable operations.  The fact that it's a little awkward in the construction case is ok in my mind currently.

include/aws/crt/ByteBuf.h
bretambrose: As general policy, should all the wrapper classes have get-implementations or explicit conversions to the c struct?  I don't have a strong opinion other than the fact that I'm going to always tend to favor a procedural approach because it's clearest to me personally.

justinboswell: I would prefer an explicit type conversion here. I don't think it's helpful to users for the most part, because the C buffer (in this case) is harder to use than the C++ one, but it makes our code a little less verbose when jumping over the API barrier.

My $0.02.

include/aws/crt/ByteBuf.h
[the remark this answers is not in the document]
justinboswell: There are also cases where API calls require a buffer, even though we might be borrowing the memory or using stack memory or something.

include/aws/crt/ByteBuf.h
```diff
@@ -0,0 +1,109 @@
...
+            // All of non-init APIs go here
+            AwsCrtResult<void> Append(ByteCursor cursor) noexcept;
+            AwsCrtResult<void> AppendDynamic(ByteCursor cursor) noexcept;
+            // etc...
+
+            aws_byte_buf *GetImpl() noexcept { return m_bufferPtr; }
+            const aws_byte_buf *GetImpl() const noexcept { return m_bufferPtr; }
+
+            const uint8_t *GetBuffer() const noexcept { return m_bufferPtr->buffer; }
+            size_t GetLen() const noexcept { return m_bufferPtr->len; }
```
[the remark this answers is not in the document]
justinboswell: +1

include/aws/crt/ByteBuf.h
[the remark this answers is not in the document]
justinboswell: I agree with Bret here. I'm not against iterators at all, and I think it would be a great thing for customers, but I want to see what the users actually do first. It may turn out they take the buffers and jam them into their own vectors or other structures immediately, then iterate.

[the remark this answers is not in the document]
marcomagdy: Granted. But I worry if you give them ByteCursor, they'll use it, even if there's a better approach lurking.
To get the same results you're mentioning would be to introduce iterators and see if it's not enough, then you would introduce a bytecursor. If you start with bytecursor, you will never have a need for iterators.

include/aws/crt/ByteBuf.h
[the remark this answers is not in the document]
marcomagdy: > Passing clib-created byte bufs up to the SDK. For example, a callback where the user is supposed to write outgoing data to an internal byte buf.

Wouldn't the user then write to the iterator instead? sorry if I'm missing the point again.

include/aws/crt/ByteBuf.h
marcomagdy: >I don't see how failing assignment could be handled with that approach.

I don't see how the current approach handles failing assignment.

bretambrose: Can't fail what you don't allow =p

bretambrose:
On second thought, let's not.

PR #60, opened 2019-09-20
Fixed memory leak in Http Stream, updated dependencies and made relev…

JonathanHenson:
…ant api compat updates.


source/io/Stream.cpp
```diff
@@ -92,6 +92,7 @@ static int s_aws_input_stream_cpp_get_length(struct aws_input_stream *stream, in
 static void s_aws_input_stream_cpp_clean_up(struct aws_input_stream *stream)
```
bretambrose: Can you rename this and the vtable entry _destroy? Since that's what they got renamed to in aws-c-io

PR #77, opened 2019-11-27
Fix MSVC compiler warnings

EeroKurimo:
*Description of changes:*

Building the project with Visual Studio 2019 currently fails due to compiler warnings treated as errors, and these fixes should address them.

---

### DIFF
--- a/tests/IotServiceTest.cpp
+++ b/tests/IotServiceTest.cpp
@@ -13,10 +13,6 @@
  * permissions and limitations under the License.
  */
 
-#ifdef _WIN32
-#define _CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES 1
-#endif
-
 #include <aws/crt/Api.h>
 
 #include <aws/testing/aws_test_harness.h>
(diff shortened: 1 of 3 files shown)


EeroKurimo:
Bummer, a failed test. It worked on my machine!

I don't have access to the failure log, would you able to share it, or maybe even just modify my PR if it's something super trivial?

JonathanHenson:
It’s probably the format check on the clang6 run. Just run clang-format -i <each file you changed> I’ll get you logs as soon as i can log into our aws account

tests/IotServiceTest.cpp
```diff
@@ -12,6 +12,11 @@
  * express or implied. See the License for the specific language governing
  * permissions and limitations under the License.
  */
+
+#ifdef _WIN32
```
JonathanHenson: Is this compiler or windows related? My guess is you want to check for msvc not windows....

EeroKurimo: It was actually compiler related, so ``_MSC_VER`` would have been the proper define to check.

However I started thinking that ``snprintf`` also satisfies the msvc compiler, which would make the sample identical for all platforms.

EeroKurimo:
The older samples' main.cpp apparently had not been modified since you added the clang-format-6.0 tests, so I applied that to it.

EeroKurimo:
I don't have the rights merge this, so please merge it when makes sense to do it.

PR #84, opened 2019-12-04
[OPTIONAL] Convert aws-c-common from ExternalProject to subdirectory

dislogical:
Since @madebr implemented awslabs/aws-c-common#538, this is what the future could look like if we applied that kind of change to all of our projects. Eventually, We could delete `aws-common-runtime/CMakeLists.txt` and the root cmake file would just include:
```cmake
if (BUILD_DEPS)
    add_subdirectory(aws-common-runtime/aws-c-common)
    add_subdirectory(aws-common-runtime/aws-c-io)
    add_subdirectory(aws-common-runtime/aws-c-compressiont)
    add_subdirectory(aws-common-runtime/aws-c-http)
    add_subdirectory(aws-common-runtime/aws-c-mtqt)
endif()
```


ColdenCullen:
TBD on how this would work with cmake import paths, apparently

kankri:
I hadn't noticed awslabs/aws-c-common#538 but have been wanting to do the same thing for the subprojects, so that they could be fetched with [`FetchContent_MakeAvailable()`](https://cliutils.gitlab.io/modern-cmake/chapters/projects/fetch.html) instead of being downloaded, built and installed in the configuration phase like today.

I'm thinking that using `FetchContent_MakeAvailable()` would be easier to customize and override by consumers of `aws-crt-cpp` than using Git submodules.

JonathanHenson:
I’d rather do them all at once, and use submodules if we go this route. Thoughts?

ColdenCullen:
> I'm thinking that using `FetchContent_MakeAvailable()` would be easier to customize and override by consumers of `aws-crt-cpp` than using Git submodules.

I was not aware of `FetchContent_MakeAvailable()`, but that definitely seems cool! I still don't fully understand it, but I'd love to see a mockup of how it could work.

> I’d rather do them all at once, and use submodules if we go this route. Thoughts?

Absolutely, I don't think we should merge this as is. I just did it as an example of how we could simplify our code if we put the time into fixing our projects' cmake scripts.

EeroKurimo:
The rough idea is that[ `FetchContent_MakeAvailable`](https://cliutils.gitlab.io/modern-cmake/chapters/projects/fetch.html) downloads the external dependency during the cmake configure step, and adds them to be compiled during the cmake build step.

The `FetchContent_MakeAvailable` requires CMake 3.14, although starting from CMake 3.11 you could already achieve the same by first calling  `FetchContent_Populate` to download the dependency during the configuration step, and then calling `add_subdirectory` to compile it during the build step.

kankri:
Yes, `FetchContent_MakeAvailable` requires CMake 3.14, but like shown in the linked document, it can be transparently emulated with a macro since CMake 3.11:

```
if(${CMAKE_VERSION} VERSION_LESS 3.14)
    macro(FetchContent_MakeAvailable NAME)
        FetchContent_GetProperties(${NAME})
        if(NOT ${NAME}_POPULATED)
            FetchContent_Populate(${NAME})
            add_subdirectory(${${NAME}_SOURCE_DIR} ${${NAME}_BINARY_DIR})
        endif()
    endmacro()
endif()
```
@JonathanHenson:
> I’d rather ..., and use submodules if we go this route. Thoughts?

When I'm using `aws-crt-cpp`, i have the following dependecy tree:
* "my app"
   *  `aws-crt-cpp`
      *  `aws-c-common`
      * `aws-c-io`
      * `aws-c-mqtt`
      * `aws-c-auth`
      * `aws-c-http`

I might want to "customize" the SDK in several ways:
* I might want to override the version of e,g, `aws-c-common` to earlier or later to work around a bug
* I might want to take e.g. `aws-c-io` from a local fork which includes bug fixes
* I might want to take all of the modules from an in-house mirror of the official repositories

When using `FetchContent_MakeAvailable` I can do any of these just by overriding some variables (e.g. `AWS_C_COMMON_SHA` or `AWS_C_IO_URL`) in `CMakeLists.txt` of "my app". With Git submodules in practice I'd have to fork `aws-crt-cpp` and edit its submodule definitions and then keep that fork updated with the origin.

JonathanHenson:
this is a work in progress, but here ya go!:

https://github.com/awslabs/aws-crt-cpp/tree/move_to_submodules

kankri:
@JonathanHenson: I see that you decided to keep pursuing the Git submodules approach. Can you comment how I could achieve the 3 example use cases I mentioned earlier which would be easy to do if CMake was used to manage the configuration of the project?

justinboswell:
> @JonathanHenson: I see that you decided to keep pursuing the Git submodules approach. Can you comment how I could achieve the 3 example use cases I mentioned earlier which would be easy to do if CMake was used to manage the configuration of the project?

A fork is exactly the way to manage this. You can make local changes, and update from upstream at a time of your choosing, and choose which fixes/commits to merge. Due to all the use cases we have to support both internally and externally, using ExternalProject or FetchContent was not really viable.

PR #89, opened 2019-12-31
Implemented a C++ based stream that will interoperate with the low-le…

JonathanHenson:
…vel aws_input_stream apis.


include/aws/crt/io/Stream.h
```diff
@@ -16,21 +16,110 @@
...
+
+                explicit operator bool() const noexcept { return IsGood(); }
+
+                virtual bool IsGood() const noexcept;
+
+                aws_input_stream *GetUnderlyingStream() noexcept { return m_underlying_stream; }
+
+              protected:
+                Allocator *m_allocator;
+                aws_input_stream *m_underlying_stream;
```
graebm: you could store `aws_input_stream m_underlying_stream` by value and avoid an allocation at construction time

source/io/Stream.cpp
```diff
@@ -13,132 +13,172 @@
...
+                buffer.len += static_cast<size_t>(m_stream->gcount());
+
+                return true;
+            }
+
+            StreamStatus StdIOStreamInputStream::GetStatus() const noexcept
+            {
+                StreamStatus status;
+                status.is_end_of_stream = m_stream->eof();
+                status.is_valid = m_stream->good();
```
graebm: Not 100% sure, but shouldn't you use `m_stream->bool()` instead of `m_stream->good()`?

Check out the graph at the bottom of [this page](https://en.cppreference.com/w/cpp/io/ios_base/iostate). The stream is no longer "good" once you get to eos, but is still "bool() == true".

PS: I am forever going to reference that graph when I hear we should handle errors like how std::iostream does it.

source/io/Stream.cpp
```diff
@@ -13,132 +13,172 @@
...
+                switch (seekBasis)
+                {
+                    case SeekBasis::Begin:
+                        seekDir = std::ios_base::beg;
+                        break;
+                    case SeekBasis::End:
+                        seekDir = std::ios_base::end;
+                        break;
+                    default:
+                        return false;
```
graebm: returning false without raising error

source/io/Stream.cpp
```diff
@@ -13,132 +13,172 @@
...
-                AWS_ZERO_STRUCT(*impl);
+                m_stream->seekg(currentPosition);
+
+                return retVal;
+            }
 
-                input_stream->allocator = allocator;
-                input_stream->vtable = &s_aws_input_stream_cpp_vtable;
-                input_stream->impl = impl;
+            bool StdIOStreamInputStream::Seek(OffsetType offsetType, SeekBasis seekBasis) noexcept
```
graebm: trivial: param should probably just be named `offset`, not `offsetType`

source/io/Stream.cpp
```diff
@@ -13,132 +13,172 @@
...
+            {
+                auto currentPosition = m_stream->tellg();
+
+                m_stream->seekg(0, std::ios_base::end);
+                int64_t retVal = -1;
+
+                if (m_stream->good())
                 {
-                    return NULL;
+                    retVal = static_cast<int64_t>(m_stream->tellg());
```
graebm: again, I really don't know if you should be checking `good()` vs `operator bool()` because `good()` will be false once we've read to EOF (though I don't *think* seeking to end will set the eof flag?)

source/io/Stream.cpp
```diff
@@ -13,132 +13,172 @@
...
+                return retVal;
+            }
 
-                input_stream->allocator = allocator;
-                input_stream->vtable = &s_aws_input_stream_cpp_vtable;
-                input_stream->impl = impl;
+            bool StdIOStreamInputStream::Seek(OffsetType offsetType, SeekBasis seekBasis) noexcept
+            {
+                // very important, otherwise the stream can't be reused after reading the entire stream the first time.
+                m_stream->clear();
```
graebm: Don't think this is necessary?
From [seekg docs](https://en.cppreference.com/w/cpp/io/basic_istream/seekg)
> Before doing anything else, seekg clears eofbit. | (since C++11)

source/io/Stream.cpp
```diff
@@ -13,132 +13,172 @@
...
-                    sizeof(struct aws_input_stream),
-                    &impl,
-                    sizeof(struct aws_input_stream_cpp_impl));
-
-                if (!input_stream)
+            }
+
+            bool StdIOStreamInputStream::Read(ByteBuf &buffer) noexcept
+            {
+                m_stream->read(reinterpret_cast<char *>(buffer.buffer + buffer.len), buffer.capacity - buffer.len);
```
graebm: I think you want to be using [readsome()](https://en.cppreference.com/w/cpp/io/basic_istream/readsome):
> Extracts up to count immediately available characters from the input stream

instead of [read()](https://en.cppreference.com/w/cpp/io/basic_istream/read):
>Characters are extracted and stored until any of the following conditions occurs:
>  * count characters were extracted and stored 
>  * end of file condition occurs on the input sequence (in which case, setstate(failbit|eofbit) is called). The number of successfully extracted characters can be queried using gcount(). 

also, readsome() returns the number of characters read, you don't need to call `gcount()` afterwards

source/io/Stream.cpp
```diff
@@ -13,132 +13,172 @@
...
+                buffer.len += static_cast<size_t>(m_stream->gcount());
+
+                return true;
```
graebm: should be be checking `good()`/`bool()` after reading? I genuinely don't know

source/io/Stream.cpp
JonathanHenson: It’s necessary, this is a day long but tracking result.

graebm: So is the quote I pulled from cppreference not correct? Or is not correctly implemented on some specific platform? I'm just curious

include/aws/crt/io/Stream.h
```diff
@@ -16,21 +16,110 @@
...
+                virtual ~InputStream();
+
+                InputStream(const InputStream &) = delete;
+                InputStream &operator=(const InputStream &) = delete;
+                InputStream(InputStream &&) = delete;
+                InputStream &operator=(InputStream &&) = delete;
+
+                explicit operator bool() const noexcept { return IsGood(); }
+
+                virtual bool IsGood() const noexcept;
```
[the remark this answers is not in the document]
JonathanHenson: Well, the base class has to do setup stuff, how about I just leave m_good protected and unmark the function as virtual?

include/aws/crt/io/Stream.h
```diff
@@ -16,21 +16,110 @@
...
+
+                /**
+                 * Seek's the stream to seekBasis based offsetType bytes.
+                 *
+                 * It is expected, that if seeking to the beginning of a stream,
+                 * all error's are cleared if possible.
+                 *
+                 * @return true on success, false otherwise.
+                 */
+                virtual bool Seek(OffsetType offsetType, SeekBasis seekBasis) noexcept = 0;
```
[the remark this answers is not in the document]
JonathanHenson: I like this idea.

source/io/Stream.cpp
```diff
@@ -13,132 +13,172 @@
...
 
-    auto currentPosition = impl->stream->tellg();
+                if (length != -1)
+                {
+                    *out_length = length;
+                    return AWS_OP_SUCCESS;
+                }
 
-    impl->stream->seekg(0, std::ios_base::end);
+                impl->m_good = false;
```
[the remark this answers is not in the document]
JonathanHenson: I'm not sure how we want to handle non-seekable streams yet. My guess is we don't, I've never needed them before, and aws_input_stream doesn't handle it anyways.

include/aws/crt/io/Stream.h
JonathanHenson: good idea!

source/io/Stream.cpp
JonathanHenson: it absolutely does, that's why we end up having to clear. but I'll definitely move it to the bool operator.

source/io/Stream.cpp
JonathanHenson: I just know that when it wasn't there, seek didn't work ever again. And it doesn't hurt anything to do it.

source/io/Stream.cpp
JonathanHenson: I've been debating this one, I think we should just rely on GetStatus(), as std::io_stream only returns 0 even on a failure.

JonathanHenson: The default implementations in aws-c-io reserve returning an error, to something that actually goes wrong, but not for end of stream.

source/io/Stream.cpp
JonathanHenson: sorry readsome is a giant pile of worthless shit, Have to go back to read()

source/io/Stream.cpp
JonathanHenson: actually turns out I was wrong about this, fixed.

include/aws/crt/io/Stream.h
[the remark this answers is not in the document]
graebm: The base class can't fail during setup anymore.
I'd personally leave it all up to the implementation, to avoid confusion over when it is/isn't automagically set by the base class, but that's just me.

include/aws/crt/io/Stream.h
```diff
@@ -16,21 +16,114 @@
...
+                 * Read up-to buffer::capacity - buffer::len into buffer::buffer
+                 * Increment buffer::len by the amount you read in.
+                 *
+                 * @return true on success, false otherwise. Return false, when there is nothing left to read.
+                 * You SHOULD raise an error via aws_raise_error()
+                 * if an actual failure condition occurs.
+                 * If you return false, GetStatusImpl() will be called to determine
+                 * the validity of the stream.
+                 */
+                virtual bool ReadImpl(ByteBuf &buffer) noexcept = 0;
```
graebm: why do we do check GetStatusImpl() when ReadImpl() fails, but we just set straight up set m_good=false when SeekImpl() fails?

source/io/Stream.cpp
```diff
@@ -13,132 +13,175 @@
...
-                const std::shared_ptr<Aws::Crt::Io::IStream> &stream,
-                Aws::Crt::Allocator *allocator) noexcept
+                return status.is_valid && !status.is_end_of_stream;
+            }
+
+            StreamStatus StdIOStreamInputStream::GetStatusImpl() const noexcept
+            {
+                StreamStatus status;
+                status.is_end_of_stream = m_stream->eof();
+                status.is_valid = *m_stream ? true : false;
```
graebm: trivial: be explicit `m_stream->operator bool()` or `static_cast<bool>(*m_stream)`

include/aws/crt/io/Stream.h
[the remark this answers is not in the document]
JonathanHenson: that's fine, good point. will update

include/aws/crt/io/Stream.h
```diff
@@ -16,21 +16,115 @@
...
+                virtual int64_t GetLengthImpl() const noexcept = 0;
+
+                /**
+                 * Seek's the stream to seekBasis based offset bytes.
+                 *
+                 * It is expected, that if seeking to the beginning of a stream,
+                 * all error's are cleared if possible.
+                 *
+                 * @return true on success, false otherwise. You SHOULD raise an error via aws_raise_error()
+                 * if a failure occurs. If you return false, the m_good flag will be set to false.
```
graebm: still mentioning the m_good flag

include/aws/crt/io/Stream.h
```diff
@@ -16,21 +16,115 @@
...
+                 * Read up-to buffer::capacity - buffer::len into buffer::buffer
+                 * Increment buffer::len by the amount you read in.
+                 *
+                 * @return true on success, false otherwise. Return false, when there is nothing left to read.
+                 * You SHOULD raise an error via aws_raise_error()
+                 * if an actual failure condition occurs.
+                 * If you return false, GetStatusImpl() will be called to determine
+                 * the validity of the stream.
+                 */
+                virtual bool ReadImpl(ByteBuf &buffer) noexcept = 0;
```
graebm: GetStatusImpl() is not called anymore

include/aws/crt/io/Stream.h
```diff
@@ -16,21 +16,115 @@
...
+                virtual ~InputStream();
+
+                InputStream(const InputStream &) = delete;
+                InputStream &operator=(const InputStream &) = delete;
+                InputStream(InputStream &&) = delete;
+                InputStream &operator=(InputStream &&) = delete;
+
+                explicit operator bool() const noexcept { return IsGood(); }
+
+                virtual bool IsGood() const noexcept = 0;
```
graebm: are you sure you want to publicly offer both a `bool()` and `IsGood()` function, when iostream's `good() != bool()` (good() is false when end-of-stream, but bool() is still true)

PR #104, opened 2020-03-17
Update to use stream activiation refactor in aws-c-http.

JonathanHenson:


source/http/HttpConnection.cpp
```diff
@@ -130,6 +130,7 @@ namespace Aws
                 options.socket_options = &connectionOptions.SocketOptions.GetImpl();
                 options.on_setup = HttpClientConnection::s_onClientConnectionSetup;
                 options.on_shutdown = HttpClientConnection::s_onClientConnectionShutdown;
+                options.enable_read_back_pressure = connectionOptions.EnableReadBackPressure;
```
graebm: oh, dammit. I didn't see this when we introduced in to aws-c-http. We already had a variable like this on streams. It used a different term that was closer to the terminology for HTTP/2 window stuff. It looked like this:

```
    /**
     * Set to true to manually manage the read window size.
     *
     * If this is false, the connection will maintain a constant window size.
     *
     * If this is true, the caller must manually increment the window size using aws_http_stream_update_window().
     * If the window is not incremented, it will shrink by the amount of body data received. If the window size
     * reaches 0, no further data will be received.
     */
    bool manual_window_management;
```

The effect of that variable was that, if you left it false (default) the user wouldn't have to call stream.UpdateWindow(), the connection would do it for you.

But now that we can totally turn off back-pressure, we should probably have the option on the connection, instead of the individual streams? I still like the term "manual window management" over "enable read back pressure", since that's the HTTP/2 terminology

include/aws/crt/http/HttpConnection.h
```diff
@@ -356,6 +371,8 @@ namespace Aws
                  * not be freed until the stream is completed.
                  *
                  * Returns an instance of HttpStream upon success and nullptr on failure.
+                 *
+                 * You must call HttpClientStream::Activate() to begin outgoing processing of the stream.
                  */
                 std::shared_ptr<HttpClientStream> NewClientStream(const HttpRequestOptions &requestOptions) noexcept;
```
graebm: Add a unit test that proves we clean up properly if the stream is never activated.
I know python did tricks to keep it alive from creation->on_complete and this had to change to only do tricks from activate->on_complete

source/http/HttpConnection.cpp
```diff
@@ -290,6 +280,8 @@ namespace Aws
             {
             }
 
+            HttpClientStream::~HttpClientStream() {}
```
graebm: c++ trivia: you can also write `HttpClientStream::~HttpClientStream() = default;` in the cpp file.

source/http/HttpConnection.cpp
```diff
@@ -168,41 +163,38 @@ namespace Aws
 
                 /* Do the same ref counting trick we did with HttpClientConnection. We need to maintain a reference
                  * internally (regardless of what the user does), until the Stream shuts down. */
-                auto *toSeat = static_cast<HttpClientStream *>(aws_mem_acquire(m_allocator, sizeof(HttpStream)));
+                auto *toSeat = static_cast<HttpClientStream *>(aws_mem_acquire(m_allocator, sizeof(HttpClientStream)));
```
graebm: duude. I found exactly this kind of bug in my own code today too... sizeof'ing the wrong thing but it happened to be big enough anyway

PR #107, opened 2020-03-27
Forgot to update the version number.

JonathanHenson:


JonathanHenson:
"necessary" is an overstatement, but it does help us with metrics so we can know which version everyone is using

PR #110, opened 2020-04-10
S3 Canary

rccarper:
*Description of changes:*
Adding support for the S3 Canary. Apologies for the big diffs and large PR.  For anyone who might not know, this canary is primarily intended for measuring transfer rates between EC2 and S3 (both single part and multipart), but also contains somewhat diagnostic functionality, such as measuring download rate from an HTTP Server, and a "fork mode" that divides transfers into subprocesses (these are for when your S3 numbers are low and you're trying to validate everything possible).  Some other notes:

* If you looked at the following PR a while back: https://github.com/awslabs/aws-crt-cpp/pull/87, some of this should be familiar.  Since then, there has been a lot of new code added, and the delta is still pretty big.  However, the MultipartTransferProcessor should basically be the same, as well as the S3 API operations in S3ObjectTransport.
* There are a few PR's that need to be squared away before this can be merged, so you will see some temporary submodule changes.
  * https://github.com/awslabs/aws-c-io/pull/266 (Merged)
  * https://github.com/awslabs/aws-c-io/pull/267 (Merged)
  * https://github.com/awslabs/aws-c-common/pull/620 (Merged)
* I realized while making this PR that aws-c-io has existing pipe functionality, which can be used in place of my existing OS library pipe calls.  I'll look at substituting the aws-c-io functionality in place of the those that I'm currently using.


canary/CanaryUtil.cpp
```diff
@@ -0,0 +1,34 @@
...
+#    include <sys/utsname.h>
+#endif
+
+using namespace Aws::Crt;
+
+std::string CanaryUtil::GetPlatformName()
+{
+#ifdef WIN32
+    return "windows";
+#elif unix
```
justinboswell, line 27: where is ```unix``` coming from? Why not use any of the C common platform macros, or have CMake define something for you?

canary/MeasureTransferRate.cpp
```diff
@@ -0,0 +1,484 @@
...
+#include <aws/crt/http/HttpConnectionManager.h>
+#include <aws/crt/http/HttpRequestResponse.h>
+#include <condition_variable>
+#include <mutex>
+
+using namespace Aws::Crt;
+
+namespace
+{
+    const uint64_t SinglePartObjectSize = 5ULL * 1024ULL * 1024ULL * 1024ULL;
```
justinboswell, line 33: These all feel like config options with these values as the defaults.

canary/MeasureTransferRate.cpp
```diff
@@ -0,0 +1,484 @@
...
+    if (m_canaryApp.GetOptions().sendEncrypted)
+    {
+        aws_byte_cursor serverName = ByteCursorFromCString(endpoint.c_str());
+        auto connOptions = m_canaryApp.GetTlsContext().NewConnectionOptions();
+        connOptions.SetServerName(serverName);
+        connectionManagerOptions.ConnectionOptions.TlsOptions = connOptions;
+    }
+
+    connectionManagerOptions.ConnectionOptions.Bootstrap = &m_canaryApp.GetBootstrap();
+    connectionManagerOptions.MaxConnections = 5000;
```
justinboswell, line 210: Config param?

canary/MetricsPublisher.cpp
```diff
@@ -0,0 +1,1064 @@
...
+        // Create a copy of the metrics to publish from
+        {
+            publisher->m_publishDataTaskCopy = std::move(publisher->m_publishData);
+            publisher->m_publishData = Vector<Metric>();
+        }
+    }
+
+    /* max of 20 per request */
+    Vector<Metric> metricsSlice;
+    while (!publisher->m_publishDataTaskCopy.empty() && metricsSlice.size() < 20)
```
justinboswell, line 974: Let's make this a constant, since it's a service limit (right?)

canary/S3ObjectTransport.cpp
```diff
@@ -0,0 +1,916 @@
...
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+
+            numAddresses = m_canaryApp.GetDefaultHostResolver().GetHostAddressCount(
+                m_endpoint, AWS_GET_HOST_ADDRESS_COUNT_RECORD_TYPE_A);
+
+            EmitS3AddressCountMetric(numAddresses);
+        }
+    }
+
+    // Clear our local cache in preparation for getting address from the host resolver.
```
justinboswell, line 103: Probably worth noting that the reason you throw these away is to let the host resolver manage TTLs

canary/S3ObjectTransport.cpp
```diff
@@ -0,0 +1,916 @@
...
+    const std::shared_ptr<Http::HttpRequest> &signedRequest)
+{
+    AWS_FATAL_ASSERT(conn->IsOpen());
+
+    Http::HttpRequestOptions requestOptionsToSend = requestOptions;
+    requestOptionsToSend.request = signedRequest.get();
+
+    ++m_activeRequestsCount;
+
+    // NOTE: The captures of the connection and signed request is a work around to keep these shared
```
justinboswell, line 274: Less a workaround, and more CORRECT USE OF SHARED_PTR :muscle:

format-check.sh
```diff
@@ -10,7 +10,7 @@ if NOT type $CLANG_FORMAT 2> /dev/null ; then
 fi
 
 FAIL=0
-SOURCE_FILES=`find source include tests -type f \( -name '*.h' -o -name '*.cpp' \) -not -name "cJSON.h" -not -name "cJSON.cpp"`
+SOURCE_FILES=`find source include tests samples canary -type f \( -name '*.h' -o -name '*.cpp' \) -not -name "cJSON.h" -not -name "cJSON.cpp"`
```
justinboswell, line 13: I was going to ask how the formatting was getting past clang-format, NOW I SEE THE VIOLENCE INHERENT IN THE SYSTEM

source/Api.cpp
```diff
@@ -27,6 +27,14 @@ namespace Aws
     {
         Allocator *g_allocator = Aws::Crt::DefaultAllocator();
 
+        static struct aws_log_subject_info s_crt_cpp_log_subject_infos[] = {
+            {AWS_LS_CRT_CPP_GENERAL, "CrtCppGeneral", "Subject for aws-crt-cpp logging that defies categorization."},
+            {AWS_LS_CRT_CPP_CANARY, "CrtCppCanary", "Subject for aws-crt-cpp canary logging."}};
```
justinboswell, line 32: Shouldn't this be in the canary, via a new package ID?

canary/CanaryApp.cpp
```diff
@@ -0,0 +1,351 @@
...
+      readFromParentPipe(-1), writeToParentPipe(-1), numUpTransfers(1), numUpConcurrentTransfers(0),
+      numDownTransfers(1), numDownConcurrentTransfers(0), childProcessIndex(0), measureSinglePartTransfer(false),
+      measureMultiPartTransfer(false), measureHttpTransfer(false), usingNumaControl(false), downloadOnly(false),
+      sendEncrypted(false), loggingEnabled(false), rehydrateBackup(false), forkModeEnabled(false),
+      isParentProcess(false), isChildProcess(false)
+{
+}
+
+#ifndef WIN32
+CanaryAppChildProcess::CanaryAppChildProcess() noexcept : pid(0), readFromChildPipe(-1), writeToChildPipe(-1) {}
```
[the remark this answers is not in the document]
rccarper: I think you're right--I'll try taking it out.

canary/CanaryApp.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+CanaryAppChildProcess::CanaryAppChildProcess(pid_t inPid, int32_t inReadPipe, int32_t inWritePipe) noexcept
+    : pid(inPid), readFromChildPipe(inReadPipe), writeToChildPipe(inWritePipe)
+{
+}
+#endif
+
+CanaryApp::CanaryApp(CanaryAppOptions &&inOptions) noexcept
+    : m_options(inOptions), m_apiHandle(g_allocator),
+      m_eventLoopGroup((!inOptions.isChildProcess && !inOptions.isParentProcess) ? 72 : 2, g_allocator),
```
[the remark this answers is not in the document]
rccarper: Yes, yes it should!

canary/CanaryApp.cpp
```diff
@@ -0,0 +1,351 @@
...
+    // increase it for added headroom.
+    rlimit fdsLimit;
+    getrlimit(RLIMIT_NOFILE, &fdsLimit);
+    fdsLimit.rlim_cur = 8192;
+    setrlimit(RLIMIT_NOFILE, &fdsLimit);
+#endif
+
+    // Increase channel fragment size to 256k, due to the added
+    // throughput increase.
+    const size_t KB_256 = 256 * 1024;
```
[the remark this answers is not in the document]
rccarper: Most likely.  I'll look at this one too.

canary/CanaryUtil.cpp
rccarper: It's been a few months since I wrote this part, but if I recall correctly, unix is compiler defined for all unix family OS's, and the uname sorts out which one we're actually on.  I didn't realize C common had platform macros to make this easier, or that I could use CMake for this.  I'll try to look at both!

canary/MeasureTransferRate.cpp
rccarper: Agreed!  Once we have a proper config file that we load I'll add this and the others there.  Right now everything is CLI which is already getting a bit crowded for options.

canary/MetricsPublisher.cpp
rccarper: It is!  And I will. :)

canary/S3ObjectTransport.cpp
rccarper: Ah, you're right.  I'll take out that part of the comment.

format-check.sh
rccarper: It shouldn't be though, right, since it's being added here?  (Although that samples shouldn't be there, I'll take that out.)

source/Api.cpp
rccarper: I was thinking because it exists as part of the C++ CRT code base, it was okay here.  But yeah, I agree, it's probably cleaner inside of the canary itself given it's a separate executable.

canary/MeasureTransferRate.cpp
rccarper: Probably should actually be based off of a handful of other constants.  I'll try to figure out a better way to define this.

canary/S3ObjectTransport.cpp
rccarper: It's actually because we're building up our own cache at this point to divide them among multiple connection managers.  At some point we should be able to do away with some of this setup and hopefully just have the host resolver handle a lot of this.

PR #111, opened 2020-04-28
Elasticurl

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- elasticurl
- add h2 unittest to HttpClientTest
- TODO: We still use command_line_parser from C land. C++ cli_parser probably will be added to the TODO list. 


bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+#include <mutex>
+
+using namespace Aws::Crt;
+
+#define ELASTICURL_VERSION "0.0.1"
+
+std::shared_ptr<Aws::Crt::Io::IStream> input_body = nullptr;
+std::ifstream inputfile;
+
+struct elasticurl_ctx
```
TingDaoK, line 36: Should it be converted to a class object?

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+    const char *capath;
+    const char *cert;
+    const char *key;
+    int connect_timeout;
+    const char *header_lines[10];
+    size_t header_line_count;
+    const char *alpn;
+    bool include_headers;
+    bool insecure;
+    FILE *output;
```
TingDaoK, line 52: ofstream?

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+#define ELASTICURL_VERSION "0.0.1"
+
+std::shared_ptr<Aws::Crt::Io::IStream> input_body = nullptr;
+std::ifstream inputfile;
+
+struct elasticurl_ctx
+{
+    struct aws_allocator *allocator;
+    const char *verb;
+    struct aws_uri uri;
```
TingDaoK, line 40: Aws::Crt::Io::Uri ?

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+    bool insecure;
+    FILE *output;
+    const char *trace_file;
+    Aws::Crt::LogLevel log_level;
+    enum aws_http_version required_http_version;
+    bool exchange_completed;
+    bool bootstrap_shutdown_completed;
+};
+
+static void s_usage(int exit_code)
```
TingDaoK, line 60: I removed all those signing-lib and signing-func things, since I don't know what they are...... Should I put them back?

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+            }
+            case 'g':
+            {
+                inputfile.open(aws_cli_optarg, std::ios::in);
+                if (!inputfile.is_open())
+                {
+                    fprintf(stderr, "unable to open file %s.\n", aws_cli_optarg);
+                    s_usage(1);
+                }
+                input_body = std::make_shared<std::ifstream>(inputfile);
```
TingDaoK, line 169: Compile Error here. However, I don't know how to fix it.....

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+        tlsConnectionOptions = tlsContext.NewConnectionOptions();
+
+        tlsConnectionOptions.SetServerName(hostName);
+        tlsConnectionOptions.SetAlpnList(app_ctx.alpn);
+    }
+    else
+    {
+        if (app_ctx.required_http_version == AWS_HTTP_VERSION_2)
+        {
+            fprintf(stderr, "Error, we don't support h2c, please use TLS for HTTP2 connection");
```
TingDaoK, line 346: should all those fprintf be converted to c++ style? std::cerr<<?

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+
+    auto stream = connection->NewClientStream(requestOptions);
+    stream->Activate();
+
+    semaphore.wait(semaphoreULock, [&]() { return streamCompleted; });
+
+    connection->Close();
+    semaphore.wait(semaphoreULock, [&]() { return connectionShutdown; });
+
+    /* TODO: Wait until the bootstrap finishing shutting down, we may need to break the API for create the Bootstrap, a
```
TingDaoK, line 524: Right now, the bootstrap on_shutdown_complete callback is always NULL, which may need API change, and it may be the follow up PR or I can try in this PR.

bin/elasticurl/main.cpp
TingDaoK: And I ignored some error handling for now, I'll fix them as I finish this up

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+    {"verbose", AWS_CLI_OPTIONS_REQUIRED_ARGUMENT, NULL, 'v'},
+    {"version", AWS_CLI_OPTIONS_NO_ARGUMENT, NULL, 'V'},
+    {"http2", AWS_CLI_OPTIONS_NO_ARGUMENT, NULL, 'w'},
+    {"http1_1", AWS_CLI_OPTIONS_NO_ARGUMENT, NULL, 'W'},
+    {"help", AWS_CLI_OPTIONS_NO_ARGUMENT, NULL, 'h'},
+    /* Per getopt(3) the last element of the array has to be filled with all zeros */
+    {NULL, AWS_CLI_OPTIONS_NO_ARGUMENT, NULL, 0},
+};
+
+static void s_parse_options(int argc, char **argv, struct elasticurl_ctx *ctx)
```
graebm, line 116: You **basically never** pass arguments by raw pointer in c++. Usually you pass by reference, ex: `ElasticurlCtx &ctx`. A reference and a pointer are the same thing as far as the compiled code is concerned, but a reference can never be NULL.
The only times you'd still use raw pointers are:
- If the argument is OPTIONAL
- `**argv` because it's as old as the dinosaurs
- `const char *` for non-owning strings (c++17 introduced std::string_view, but we're limiting ourselves to c++14)

Oh yeah, and you don't need to write `struct` except where the struct is initially declared.

bin/elasticurl/main.cpp
graebm: sure
honestly C++ programmers often use printf over std::cerr, but I think in this case std::cerr will be simpler

dont forget to end lines wih "\n" or `<< std::endl;`

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,540 @@
...
+    (void)argc;
+    (void)argv;
+
+    struct elasticurl_ctx app_ctx;
+    AWS_ZERO_STRUCT(app_ctx);
+    app_ctx.allocator = allocator;
+    app_ctx.connect_timeout = 3000;
+    app_ctx.output = stdout;
+    app_ctx.verb = "GET";
+    app_ctx.alpn = "h2;http/1.1";
```
graebm, line 284: Just set the initial values for members in the struct definition. Don't zero it out here

source/http/HttpConnection.cpp
```diff
@@ -205,6 +205,8 @@ namespace Aws
 
             void HttpClientConnection::Close() noexcept { aws_http_connection_close(m_connection); }
 
+            enum aws_http_version HttpClientConnection::GetVersion() noexcept { return aws_http_connection_get_version(m_connection); }
```
graebm, line 208: Declare this enum in C++
```
enum class HttpVersion {
    Unknown = AWS_HTTP_VERSION_UNKNOWN,
    Http1_1 = AWS_HTTP_VERSION_1_1,
    ...
};
```

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,522 @@
...
+    Io::TlsContextOptions tlsCtxOptions;
+    Io::TlsContext tlsContext;
+    Io::TlsConnectionOptions tlsConnectionOptions;
+    if (use_tls)
+    {
+        if (appCtx.cert && appCtx.key)
+        {
+            tlsCtxOptions = Io::TlsContextOptions::InitClientWithMtls(appCtx.cert, appCtx.key);
+        }
+        else
```
TingDaoK, line 299: We got an `#ifdef _WIN32` branch here in c elasticurl. But I cannot find the c++ binding for `aws_tls_ctx_options_init_client_mtls_from_system_path`.

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,523 @@
...
+    const char *alpn = "h2;http/1.1";
+    bool IncludeHeaders = false;
+    bool insecure = false;
+
+    const char *TraceFile = nullptr;
+    LogLevel LogLevel = LogLevel::None;
+    Http::HttpVersion RequiredHttpVersion = Http::HttpVersion::Unknown;
+
+    std::shared_ptr<Io::IStream> InputBody = nullptr;
+    std::ofstream OutPut;
```
graebm, line 53: [trivial] don't need to capitalize P, it's all one word

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    Http::HttpHeader userAgentHeader;
+    userAgentHeader.name = ByteCursorFromCString("user-agent");
+    userAgentHeader.value = ByteCursorFromCString("elasticurl_cpp 1.0, Powered by the AWS Common Runtime.");
+    request.AddHeader(userAgentHeader);
+
+    std::shared_ptr<Io::StdIOStreamInputStream> bodyStream =
+        MakeShared<Io::StdIOStreamInputStream>(allocator, appCtx.InputBody, allocator);
+    int64_t dataLen = 0;
+    if (aws_input_stream_get_length(bodyStream->GetUnderlyingStream(), &dataLen))
```
graebm, line 479: we shouldn't need to resort to the C API.
Let's either:
1) get the length from appCtx.InputBody
2) Add the public API to Aws::Crt::Io::InputStream: Seek(), Read(), GetStatus(), GetLength()

bin/elasticurl/main.cpp
```diff
@@ -0,0 +1,523 @@
...
+    {
+        std::cerr << "failed to get length of input stream.\n";
+        exit(1);
+    }
+    if (dataLen > 0)
+    {
+        char contentLength[64];
+        AWS_ZERO_ARRAY(contentLength);
+        sprintf(contentLength, "%li", dataLen);
```
graebm, line 488: std::string contentLength = std::to_string(dataLen);

bin/elasticurl/main.cpp
TingDaoK: did the second option

bin/elasticurl_cpp/main.cpp
```diff
@@ -0,0 +1,552 @@
...
+        std::cerr << "Failed to create client bootstrap with error " << aws_error_debug_str(clientBootstrap.LastError())
+                  << std::endl;
+        exit(1);
+    }
+    clientBootstrap.EnableBlockingShutdown();
+
+    std::promise<std::shared_ptr<Http::HttpClientConnection>> connectionPromise;
+    std::promise<void> shutdownPromise;
+
+    auto onConnectionSetup = [&](const std::shared_ptr<Http::HttpClientConnection> &newConnection, int errorCode) {
```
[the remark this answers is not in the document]
TingDaoK: Sorry, I am not sure what's not necessary here.

bin/elasticurl_cpp/main.cpp
```diff
@@ -0,0 +1,552 @@
...
+    bool IncludeHeaders = false;
+    bool insecure = false;
+
+    const char *TraceFile = nullptr;
+    Aws::Crt::LogLevel LogLevel = Aws::Crt::LogLevel::None;
+    Http::HttpVersion RequiredHttpVersion = Http::HttpVersion::Unknown;
+
+    std::shared_ptr<Io::IStream> InputBody = nullptr;
+    std::ofstream Output;
+};
```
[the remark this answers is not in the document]
TingDaoK: yep, will do

bin/elasticurl_cpp/main.cpp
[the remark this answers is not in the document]
graebm: David's saying you should be explicit and name the exact variables used in the lambda capture, like:
`[&appCtx, &connectionPromise]` instead of just `[&]`

Then he was warning that clang will yell at you if you list more variables than are actually used

bin/elasticurl_cpp/main.cpp
```diff
@@ -0,0 +1,552 @@
...
+    request.SetPath(appCtx.uri.GetPathAndQuery());
+
+    Http::HttpHeader hostHeader;
+    hostHeader.name = ByteCursorFromCString("host");
+    hostHeader.value = appCtx.uri.GetHostName();
+    request.AddHeader(hostHeader);
+
+    Http::HttpHeader userAgentHeader;
+    userAgentHeader.name = ByteCursorFromCString("user-agent");
+    userAgentHeader.value = ByteCursorFromCString("elasticurl_cpp 1.0, Powered by the AWS Common Runtime.");
```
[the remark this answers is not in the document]
graebm: Eh, this matches what's in the other Elasticurl's, and this whole thing is a toy anyway. I wouldn't stress

PR #112, opened 2020-04-29
Bootstrap shutdown

graebm:
Allow user to know when `ClientBootstrap` is done shutting down its behind-the-scenes resources.

User may register a `std::function<void()>` that fires upon shutdown complete. I considered offering a `std::future` to users, but realized you can only do blocking wait()/get() operations on a c++ future, you can't register a callback to fire (a feature available in Python and Java's future primitives). So if a user wants to use `std::future` they can hook that up themselves with the callback (I did this in a test).

Also added an `EnableBlockingShutdown()` function. I was looking around at how async-shutdown was handled in other `aws-crt-cpp` classes and saw this feature on the `HttpClientConnectionManager`. It's simplifies things when the IO classes are created on the main thread's stack (all of our tests and samples).

I chose to make these functions you call on the constructed class, rather than constructor arguments, or fields in an "Options" struct for 2 reasons: 1) This class wasn't using an Options struct before, so it felt weird to add one now. 2) If these were overloaded constructor args, I feared a user would enable the "blocking shutdown" by copy/pasting a constructor with `..., true, ...` in it and later have no idea why they occasionally dead-locked.

### DIFF
--- a/include/aws/crt/io/Bootstrap.h
+++ b/include/aws/crt/io/Bootstrap.h
@@ -30,6 +30,17 @@ namespace Aws
     {
         namespace Io
         {
+            using OnClientBootstrapShutdownComplete = std::function<void()>;
+
+            /**
+             * A ClientBootstrap handles creation and setup of socket connections
+             * to specific endpoints.
+             *
+             * Note that ClientBootstrap may not clean up all its behind-the-scenes
+             * resources immediately upon destruction. If you need to know when
+             * behind-the-scenes shutdown is complete, use SetShutdownCompleteCallback()
+             * or EnableBlockingShutdown() (only safe on main thread).
+             */
             class AWS_CRT_CPP_API ClientBootstrap final
             {
               public:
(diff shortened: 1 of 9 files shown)


include/aws/crt/io/Bootstrap.h
```diff
@@ -44,11 +57,36 @@ namespace Aws
                 operator bool() const noexcept;
                 int LastError() const noexcept;
 
+                /**
+                 * Set function to invoke when ClientBootstrap's behind-the-scenes
+                 * resources finish shutting down. This function may be invoked
+                 * on any thread. Shutdown begins when the ClientBootstrap's
+                 * destructor runs.
+                 */
+                void SetShutdownCompleteCallback(OnClientBootstrapShutdownComplete callback);
```
JonathanHenson, line 66: why not noexcept? move asignment is supposed to be noexcept by definition right?

source/io/Bootstrap.cpp
```diff
@@ -20,35 +20,74 @@ namespace Aws
...
                 {
+                    // Release m_callbackData, it destroys itself when shutdown completes.
+                    m_callbackData.release();
+
                     aws_client_bootstrap_release(m_bootstrap);
-                    m_bootstrap = nullptr;
-                    m_lastError = AWS_ERROR_UNKNOWN;
+                    if (m_enableBlockingShutdown)
+                    {
+                        // If your program is stuck here, stop using EnableBlockingShutdown()
```
JonathanHenson, line 74: lol

source/io/Bootstrap.cpp
```diff
@@ -20,35 +20,74 @@ namespace Aws
...
                 }
             }
 
             ClientBootstrap::operator bool() const noexcept { return m_lastError == AWS_ERROR_SUCCESS; }
 
             int ClientBootstrap::LastError() const noexcept { return m_lastError; }
 
+            void ClientBootstrap::SetShutdownCompleteCallback(OnClientBootstrapShutdownComplete callback)
+            {
+                m_callbackData->ShutdownCallback = std::move(callback);
```
JonathanHenson, line 86: I trust your judgement on this one, but shouldn't callback be an r-value?

graebm: This felt right to me. We're storing a copy, so we take this by copy (then we move it into our member).

Couldn't find a satisfying answer on Stackoverflow on the **one true way**

Ran both variants through godbolt and ended up with the same generated code so 🤷‍♀️

include/aws/crt/io/Bootstrap.h
graebm: nope, `function`'s [assignment operators](https://en.cppreference.com/w/cpp/utility/functional/function/operator%3D) are mostly excepty

PR #114, opened 2020-05-11
Docs(phase 1)

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- Generated document based on the comments we had via Doxygen
- Comments updated at crt.IO for the document.(Just IO for now. To keep consistence with aws-crt-python, I may update the MQTT for next, but let me read about MQTT frist. Also, I can base on the comment for IO to udpate MQTT) 
- We should probably create C++ enum class instead of just reuse the C ones for clarifying the doc.


include/aws/crt/Api.h
```diff
@@ -59,7 +70,14 @@ namespace Aws
         };
 
         AWS_CRT_CPP_API const char *ErrorDebugString(int error) noexcept;
+        /**
+         * @return the value of the last aws error on the current thread. Return 0 if no aws-error raised before.
+         */
         AWS_CRT_CPP_API int LastError() noexcept;
+        /**
+         * @return the value of the last aws error on the current thread.
+         * @return AWS_ERROR_UNKNOWN, if no aws-error raised before.
```
graebm, line 79: two "@return" in a comment is weird, right?

TingDaoK: Yep, removed.

PR #141, opened 2020-06-19
Add dependencies: aws-checksums and aws-c-event-stream.

wps132230:
*Issue*
[aws-sdk-cpp](https://github.com/aws/aws-sdk-cpp) depends on [aws-c-event-stream](https://github.com/awslabs/aws-c-event-stream) and [aws-chceksums](https://github.com/awslabs/aws-checksums) for event stream support.
When migrating to aws-crt-cpp, we want to have only one dependency. But aws-crt-cpp doesn't include aws-c-event-stream and aws-checksums as submodules.

*Description of changes:*
Add aws-c-event-stream and aws-checksums as another two git submodules.


wps132230:
Need to update the git submodule hash after the following PRs get merged.
So, pending merge on these two after approval.
https://github.com/awslabs/aws-c-event-stream/pull/38
https://github.com/awslabs/aws-checksums/pull/28

README.md
```diff
@@ -9,6 +9,8 @@ C++ wrapper around the aws-c-* libraries. Provides Cross-Platform Transport Prot
 * aws-c-mqtt: MQTT client.
 * aws-c-auth: Auth signers such as Aws-auth sigv4
 * aws-c-http: HTTP 1.1 client, and websockets (H2 coming soon)
+* aws-checksums: Cross-Platform HW accelerated CRC32c and CRC32 with fallback to efficient SW implementations.
```
DavidOgunsAWS, line 12: Wasn't this merged into aws-c-common?

cmake/aws-crt-cpp-config.cmake
```diff
@@ -4,6 +4,7 @@ find_dependency(aws-c-http)
 find_dependency(aws-c-mqtt)
 find_dependency(aws-c-cal)
 find_dependency(aws-c-auth)
+find_dependency(aws-c-event-stream)
```
DavidOgunsAWS, line 7: Related to other comment, if not merged into aws-c-common shouldn't aws-c-checksums be mentioned here?

wps132230: It's in aws-c-event-stream-config.cmake
https://github.com/awslabs/aws-c-event-stream/blob/master/cmake/aws-c-event-stream-config.cmake
Similar to that we don't mention aws-c-io here.
And we are not using any functions from aws-checksums directly, so, I think it's fine?

README.md
wps132230: Do you mean aws-checksums depends on aws-c-common now?

DavidOgunsAWS: Looks like you have the current state of things right.

wps132230:
Rebased on the latest master branch, and updated the git submodule hash for aws-c-event-stream and aws-checksums as the pending PRs were merged.

PR #146, opened 2020-06-30
Multiple sub bug

TingDaoK:
*Issue #, if available:*
Related with https://github.com/awslabs/aws-c-mqtt/pull/141
*Description of changes:*


TingDaoK:
Changed mind, the fix is not the BEST fix, we should still pass the struct aws_mqtt_topic_subscription to user, which is better

TingDaoK:
Still need some fix from aws-crt-cpp ....

TingDaoK:
hell no, I don't know how to fix the conflict of the submodules, just reopen another PR to make it simple.

PR #148, opened 2020-07-13
multiple sub fix from crt land

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- downstream fix for https://github.com/awslabs/aws-c-mqtt/pull/141


source/mqtt/MqttClient.cpp
```diff
@@ -181,12 +181,13 @@ namespace Aws
...
                     size_t length = aws_array_list_length(topicSubacks);
-                    Vector<String> topics(length);
+                    Vector<String> topics;
+                    topics.reserve(length);
                     QOS qos = AWS_MQTT_QOS_AT_MOST_ONCE;
                     for (size_t i = 0; i < length; ++i)
                     {
                         aws_mqtt_topic_subscription *subscription = NULL;
-                        aws_array_list_get_at_ptr(topicSubacks, reinterpret_cast<void **>(&subscription), i);
+                        aws_array_list_get_at(topicSubacks, &subscription, i);
```
TingDaoK, line 190: The array_list contains aws_mqtt_topic_subscription pointers. 
I know it's werid that the [subcribe call](https://github.com/awslabs/aws-c-mqtt/blob/master/include/aws/mqtt/client.h#L388) asks for a list of objects instead of pointers. However, the [callback](https://github.com/awslabs/aws-c-mqtt/blob/master/include/aws/mqtt/client.h#L75) will provide a list of pointers. 😑 
pros: 
- it's probably easy for user to code?
cons:
- you have to read to documentation carefully... And it's werid.

graebm: You're just fixing code that was broken before, right?
You're not changing the API are you?

TingDaoK: yeah, I just change it to make it work. Not changing any existing API.

Issue #143, opened 2020-06-23
Git version history and tag problem


kankri:
I mentioned this in https://github.com/awslabs/aws-c-mqtt/issues/135#issuecomment-648236162, but thought to create an issue to the repository which is affected.

It seems to me that existing tags v0.8.4 and v0.8.5 have been changed to point to different commits recently. My previous clone shows:

```
* commit 393433e9fcfe48ab09c02a23577724f96dcdee5f (HEAD, tag: v0.8.5)
| Author: Bret Ambrose <bretambrose@gmail.com>
| Date:   Mon Jun 15 11:31:40 2020 -0700
```

```
* commit 421a69a81fb73b4c9917ffb3d095c2c65e9e973d (tag: v0.8.4)
| Author: Justin Boswell <boswej@amazon.com>
| Date:   Mon Jun 8 15:02:10 2020 -0700
```

My current clone shows:

```
* commit 107f24676a9f99bc5c5e61abe61b69044117c8fe (tag: v0.8.4)
| Author: GitHub Actions <aws-sdk-common-runtime@amazon.com>
| Date:   Tue Jun 23 01:55:40 2020 +0000
|
|     Updated version to v0.8.4
|
* commit 8c45c499e10eb2ea49fa56698c156e9bb9583f4c (origin/update_dep)
| Author: Dengke Tang <dengket@amazon.com>
| Date:   Mon Jun 22 18:55:15 2020 -0700
* commit fb751d0ea75401423c07acd4737fd0b7a7fc2bdc (origin/master, origin/HEAD, master)
| Author: Justin Boswell <boswej@amazon.com>
| Date:   Thu Jun 18 09:09:15 2020 -0700
| * commit 31fd668ef26af0ae0ee1ee84c01f7466b023e67f (tag: v0.8.5)
| | Author: GitHub Actions <aws-sdk-common-runtime@amazon.com>
| | Date:   Wed Jun 17 23:34:12 2020 +0000
| |
| |     Updated version to v0.8.5
| |
| * commit c2dfd35805f25e28ec933a29e3a2bf77a22e5afa
|/  Author: Justin Boswell <boswej@amazon.com>
|   Date:   Wed Jun 17 23:33:37 2020 +0000
* commit 9fda1e3309a4de37c2f052c5b1d531537c70b8a3
```

The new tags don't look reasonable and existing tags shouldn't be reassigned in any case. Is this some automation running amok?

bretambrose:
[...]

PR #150, opened 2020-07-15
Only update git tags on release, not create

justinboswell:
*Issue #, if available:* https://github.com/awslabs/aws-crt-cpp/issues/143

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,8 @@
 name: Release
 
 on:
-  create
+  release:
+    types: [published]
 
 jobs:
   update-version:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,7 +1,8 @@
 name: Release
 
 on:
-  create
+  release:
+    types: [published]
```
bretambrose, line 5: Will this still work for pre-releases (dunno what published means here)?

justinboswell: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release it is unclear, but I would *hope* so. We're about to find out.

PR #151, opened 2020-07-16
X509 Credentials Provider Binding

bretambrose:
https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/129

### DIFF
--- a/include/aws/crt/auth/Credentials.h
+++ b/include/aws/crt/auth/Credentials.h
@@ -6,6 +6,8 @@
 
 #include <aws/crt/Exports.h>
 #include <aws/crt/Types.h>
+#include <aws/crt/http/HttpConnection.h>
+#include <aws/crt/io/TlsOptions.h>
 
 #include <chrono>
 #include <functional>
@@ -20,6 +22,11 @@ namespace Aws
         namespace Io
         {
             class ClientBootstrap;
+        }
+
+        namespace Http
+        {
+            class HttpClientConnectionProxyOptions;
         }
 
         namespace Auth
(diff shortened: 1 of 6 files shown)


include/aws/crt/auth/Credentials.h
```diff
@@ -6,6 +6,8 @@
 
 #include <aws/crt/Exports.h>
 #include <aws/crt/Types.h>
+#include <aws/crt/http/HttpConnection.h>
```
JonathanHenson, line 9: I don’t see where this include is being used.

source/auth/Credentials.cpp
```diff
@@ -267,6 +269,45 @@ namespace Aws
...
+                AWS_ZERO_STRUCT(proxy_options);
+                if (config.ProxyOptions.has_value())
+                {
+                    const Http::HttpClientConnectionProxyOptions &proxy_config = config.ProxyOptions.value();
+
+                    proxy_options.host = aws_byte_cursor_from_c_str(proxy_config.HostName.c_str());
+                    proxy_options.port = proxy_config.Port;
+                    proxy_options.tls_options = proxy_config.TlsOptions->GetUnderlyingHandle();
+                    proxy_options.auth_type = (enum aws_http_proxy_authentication_type)proxy_config.AuthType;
+                    proxy_options.auth_username = aws_byte_cursor_from_c_str(proxy_config.BasicAuthUsername.c_str());
```
JonathanHenson, line 296: Some of these are optional right? Should we do a check that they’ve been set?

include/aws/crt/auth/Credentials.h
bretambrose: Proxy Options is there

source/auth/Credentials.cpp
bretambrose: username and password being null will just result in 0-length cursors.  Everything else is required.

PR #163, opened 2020-08-24
IMDSv2 client Integration

singku:
*Issue #, if available:*

*Description of changes:*


source/auth/Credentials.cpp
```diff
@@ -106,7 +116,7 @@ namespace Aws
 
             CredentialsProvider::~CredentialsProvider()
             {
-                if (m_provider != nullptr)
+                if (m_provider)
```
singku, line 119: This is mainly to raise the question:
Do we need to call `aws_credentials_provider_acquire(provider)` as well in the constructor above?
To me it seems we should.

singku, line 119: This is mainly to raise the question:
Do we need to call `aws_credentials_provider_acquire(provider)` as well in the constructor above?
To me it seems we should. If we should, there are several places in this repo that needs to be cleaned up, including operations on `aws_credentials` ,`aws_credentials_provider`, etc.

bretambrose: _new() always returns an object with a ref count of 1, so there's no need to call acquire on it.

tests/IotServiceTest.cpp
```diff
@@ -74,17 +74,16 @@ static int s_TestIotPublishSubscribe(Aws::Crt::Allocator *allocator, void *ctx)
...
-                printf(
-                    "%s errorCode=%d returnCode=%d sessionPresent=%d\n",
-                    (errorCode == 0) ? "CONNECTED" : "COMPLETED",
-                    errorCode,
-                    (int)returnCode,
-                    (int)sessionPresent);
-                connected = true;
-                cv.notify_one();
-            };
+        auto onConnectionCompleted = [&](MqttConnection &, int errorCode, ReturnCode returnCode, bool sessionPresent) {
```
bretambrose, line 77: what's causing these changes?  clang-format is not making this change for me

include/aws/crt/ImdsClient.h
```diff
@@ -0,0 +1,342 @@
...
+                std::function<void(const Auth::Credentials &credentials, int errorCode, void *userData)>;
+            using OnIamProfileAcquired =
+                std::function<void(const IamProfile &iamProfile, int errorCode, void *userData)>;
+            using OnInstanceInfoAcquired =
+                std::function<void(const InstanceInfo &instanceInfo, int errorCode, void *userData)>;
+
+            class AWS_CRT_CPP_API ImdsClient
+            {
+              public:
+                ImdsClient(aws_imds_client *client, Allocator *allocator = g_allocator) noexcept;
```
bretambrose, line 98: Do we need this constructor?

source/auth/Credentials.cpp
bretambrose: I don't see anything relevant to aws_credentials but I suppose you could rework the provider construction to acquire in s_CreateWrappedProvider and release in all of the helper functions.  Anyone else using the public constructor is going to be broken though.

include/aws/crt/ImdsClient.h
```diff
@@ -0,0 +1,342 @@
...
+                /* Should add retry strategy support once that is available */
+            };
+
+            /**
+             * https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
+             */
+            struct AWS_CRT_CPP_API IamProfile
+            {
+                DateTime lastUpdated;
+                ByteCursor instanceProfileArn;
```
bretambrose, line 56: To be honest, I hate seeing byte cursors in user-facing APIs at the CRT level.

include/aws/crt/ImdsClient.h
singku: I did this following `Credentials`'s ctor. Shall we allow customer to initiate a Client using the raw client?

source/auth/Credentials.cpp
singku: Yes, that's what I encountered in the test when I tried to rework this.
What I mean is in this ctor
```
CredentialsProvider::CredentialsProvider(aws_credentials_provider *provider, Allocator *allocator) noexcept
  : m_allocator(allocator), m_provider(provider)
{	            
}	            
```
If we provide customer this ctor, without increasing ref_cnt to the raw provider. It likely will hold an invalid pointer later. There are similar cases for aws_credentials at other places. I didn't list here.

tests/IotServiceTest.cpp
singku: Clang-format did this for me though.

include/aws/crt/ImdsClient.h
bretambrose: Credentials has that constructor so that we can give something back to the user from a C callback.  With the imds client, a constructor from config is sufficient.

include/aws/crt/ImdsClient.h
bretambrose: I'm torn, maybe someone else has a good rule for when to use what.

source/auth/Credentials.cpp
bretambrose: Could you point out the similar cases because I didn't see any.

singku: After spotting the `aws_credentials_provider` issue, I did a scan of `GetUnderlyingHandle()` call and thought the following code is problematic. Now I realize that I was wrong, m_credentials is a shared_ptr, as long as it stays, m_config.credentials won't be destroyed.
```
void AwsSigningConfig::SetCredentials(const std::shared_ptr<Credentials> &credentials) noexcept
{
        m_credentials = credentials;
        m_config.credentials = m_credentials->GetUnderlyingHandle();
}
```

source/auth/Credentials.cpp
```diff
@@ -28,6 +28,16 @@ namespace Aws
                 }
             }
 
+            Credentials::Credentials(const aws_credentials *credentials, Allocator *allocator) noexcept
```
bretambrose, line 31: What is the need for this constructor?

singku: The get credentials callback from imds client provides a const aws_credentials parameter. This is used to construct a Credentials object from a const raw aws_credentials.

bretambrose: Couldn't the existing constructor have been relaxed?

include/aws/crt/ImdsClient.h
```diff
@@ -0,0 +1,341 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/crt/DateTime.h>
+#include <aws/crt/Exports.h>
+#include <aws/crt/Types.h>
+#include <functional>
+
+struct aws_array_list;
```
justinboswell, line 12: Should these be publicly named here? Couldn't we just add struct in front of their declarations in the C++ types?

source/ImdsClient.cpp
```diff
@@ -0,0 +1,401 @@
...
+                const aws_imds_instance_info *instanceInfo,
+                int errorCode,
+                void *userData)
+            {
+                WrappedCallbackArgs<OnInstanceInfoAcquired> *callbackArgs =
+                    static_cast<WrappedCallbackArgs<OnInstanceInfoAcquired> *>(userData);
+                InstanceInfo info;
+                info.marketplaceProductCodes = ArrayListToVector<ByteCursor, StringView>(
+                    &(instanceInfo->marketplace_product_codes), ByteCursorToStringView);
+                info.availabilityZone = StringView(
```
justinboswell, line 110: You have a ByteCursorToStringView function, shouldn't we use it here?

singku: We should, I forgot to make this change.

justinboswell (approved):

source/auth/Credentials.cpp
bretambrose: Another possibility (gathering feedback) is changing acquire/release to take const * and then updating everything here and in auth to be appropriately const *.

bretambrose (approved):
Looks fine, only change I'd like to see is to update aws-c-auth to include that const ref count API change as soon as its merged and then merge the two Credentials constructors so that Credentials(const struct aws_credentials *)  takes a ref to the C struct.

PR #166, opened 2020-08-27
add refcount for HttpMessage and remove ownMessage

TingDaoK:
*Issue #, if available:*
https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/146
If the underlying http_message get destroyed before the destructor of HttpMessage, it may crash when the destructor accesses the underlying http_message.

*Description of changes:*
- Keep the underlying http_message alive until the destructor called


TingDaoK:
> I'd prefer to remove ownMessage entirely.
> 
> Always acquire it, always release it, and in the HttpRequest::HttpRequest(Allocator *allocator) constructor, dec ref the message in the body.

Yeah. Totally. I also feel weird about the ownMessage... Thanks for the advice! Will update soon!

PR #169, opened 2020-09-02
Support string_view

singku:
*Issue #, if available:*

*Description of changes:*


include/aws/crt/StringView.h
```diff
@@ -0,0 +1,136 @@
...
+            typedef std::reverse_iterator<const_iterator> const_reverse_iterator;
+            typedef const_reverse_iterator reverse_iterator;
+            static constexpr const size_t npos = static_cast<size_t>(-1);
+
+            StringView() noexcept : m_size(0), m_data(nullptr) {}
+            StringView(const StringView &other) noexcept = default;
+            StringView(const char *s, size_t size) noexcept : m_size(size), m_data(s) {}
+            StringView(const char *s) noexcept : m_size(std::char_traits<char>::length(s)), m_data(s) {}
+
+            StringView &operator=(StringView &other) noexcept = default;
```
graebm, line 40: trivial: **const** StringView &other

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,136 @@
...
+            typedef const_pointer const_iterator;
+            typedef const_iterator iterator;
+            typedef std::reverse_iterator<const_iterator> const_reverse_iterator;
+            typedef const_reverse_iterator reverse_iterator;
+            static constexpr const size_t npos = static_cast<size_t>(-1);
+
+            StringView() noexcept : m_size(0), m_data(nullptr) {}
+            StringView(const StringView &other) noexcept = default;
+            StringView(const char *s, size_t size) noexcept : m_size(size), m_data(s) {}
+            StringView(const char *s) noexcept : m_size(std::char_traits<char>::length(s)), m_data(s) {}
```
graebm, line 38: These 2 constructors are NOT noexcept in the [standard](https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view) (don't know why though)

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,136 @@
...
+            typedef const char *const_pointer;
+            typedef char &reference;
+            typedef const char &const_reference;
+            typedef const_pointer const_iterator;
+            typedef const_iterator iterator;
+            typedef std::reverse_iterator<const_iterator> const_reverse_iterator;
+            typedef const_reverse_iterator reverse_iterator;
+            static constexpr const size_t npos = static_cast<size_t>(-1);
+
+            StringView() noexcept : m_size(0), m_data(nullptr) {}
```
graebm, line 35: Seems like every constructor and function on string_view is constexpr in the [standard](https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view). Can we do that with c++11?

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,136 @@
...
+
+            size_t size() const noexcept { return m_size; }
+
+            size_t length() const noexcept { return m_size; }
+
+            size_t max_size() const noexcept { return (std::numeric_limits<size_t>::max)(); }
+
+            bool empty() const noexcept { return m_size == 0; }
+
+            const_reference operator[](size_t pos) const noexcept
```
graebm, line 66: operator[] is NOT noexcept [link](https://en.cppreference.com/w/cpp/string/basic_string_view/operator_at)

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,136 @@
...
+            {
+                assert(pos < m_size);
+                return m_data[pos];
+            }
+
+            const_reference at(size_t pos) const noexcept
```
graebm, line 72: at is NOT noexcept

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,136 @@
...
+                return m_data[pos];
+            }
+
+            const_reference front() const noexcept
+            {
+                assert(m_size != 0);
+                return m_data[0];
+            }
+
+            const_reference back() const noexcept
```
graebm, line 84: front and back are not noexcept

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,136 @@
...
+            const_pointer data() const noexcept { return m_data; }
+
+            void remove_prefix(size_t n) noexcept
+            {
+                assert(n <= m_size);
+                m_data += n;
+                m_size -= n;
+            }
+
+            void remove_suffix(size_t n) noexcept
```
graebm, line 99: remove_prefix/suffix are NOT noexcept

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,136 @@
...
+                assert(pos <= m_size);
+                return StringView(m_data + pos, (std::min)(n, m_size - pos));
+            }
+
+          private:
+            size_t m_size;
+            const char *m_data;
+        };
+#endif
+    } // namespace Crt
```
graebm, line 135: There's other stuff we haven't defined.
Definitely useful:
- comparison functions (operator==, etc)
- std::hash

Possibly useful:
- operator<<
- find, rfind, find_first_of, find_last_of, find_first_not_of, find_last_not_of
- member types: traits_type, value_type, pointer, const_pointer, etc...

include/aws/crt/StringView.h
singku: Those NOT noexcept functions throw in standard implementation. Here in crt-cpp we don't use throw.

tests/StringViewTest.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+static int s_test_string_view(struct aws_allocator *allocator, void *ctx)
+{
+    (void)ctx;
+    Aws::Crt::ApiHandle apiHandle(allocator);
+
+    {
+        const char *data = "abc123xyz";
+
+        Aws::Crt::StringView sv(data);
```
bretambrose, line 18: While these don't need to be split up as individual tests, it would be nice if they were at least broken up by subject/theme.

include/aws/crt/external/StringView.h
```diff
@@ -0,0 +1,819 @@
...
+    {
+        size_t operator()(const basic_string_view<_CharT, _Traits> __val) const noexcept;
+    };
+
+    template <class _CharT, class _Traits>
+    size_t hash<basic_string_view<_CharT, _Traits>>::operator()(const basic_string_view<_CharT, _Traits> __val) const
+        noexcept
+    {
+        auto str = std::basic_string<_CharT, _Traits>(__val.data(), __val.size());
+        return std::hash<std::basic_string<_CharT, _Traits>>()(str);
```
singku, line 815: This is not ideal, even strange when using string in string_view. But to keep the hash of a basic_string and basic_string_view the same when underlying char sequence are the same, this is the best choice I can think of currently.

PR #171, opened 2020-09-04
Ref

bretambrose:
* Updates CRT to use new ref-counting APIs for a number of C types

### DIFF
--- a/source/io/EventLoopGroup.cpp
+++ b/source/io/EventLoopGroup.cpp
@@ -20,14 +20,7 @@ namespace Aws
                 }
             }
 
-            EventLoopGroup::~EventLoopGroup()
-            {
-                if (!m_lastError)
-                {
-                    aws_event_loop_group_release(m_eventLoopGroup);
-                    m_lastError = AWS_ERROR_SUCCESS;
-                }
-            }
+            EventLoopGroup::~EventLoopGroup() { aws_event_loop_group_release(m_eventLoopGroup); }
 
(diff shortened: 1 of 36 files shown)


source/io/EventLoopGroup.cpp
```diff
@@ -25,25 +24,25 @@ namespace Aws
             {
                 if (!m_lastError)
                 {
-                    aws_event_loop_group_clean_up(&m_eventLoopGroup);
+                    aws_event_loop_group_release(m_eventLoopGroup);
```
singku, line 27: Though `m_lastError` currently only refers to the `new` operation result, better to use
```
if (m_eventLoopGroup)
{
    release
}
```
or simply
```
release
```
if `aws_event_loop_group_release` has the protection.
in case later we add more operations and m_lastError doesn't necessarily identify the existence of underlying handle.

source/io/TlsOptions.cpp
```diff
@@ -261,19 +261,17 @@ namespace Aws
                 if (mode == TlsMode::CLIENT)
                 {
                     aws_tls_ctx *underlying_tls_ctx = aws_tls_client_ctx_new(allocator, &options.m_options);
-
                     if (underlying_tls_ctx != NULL)
                     {
-                        m_ctx.reset(underlying_tls_ctx, aws_tls_ctx_destroy);
+                        m_ctx.reset(underlying_tls_ctx, aws_tls_ctx_release);
```
singku, line 266: nit: ` != nullptr`

justinboswell (approved):

PR #174, opened 2020-09-16
Fix string_view in C++ 17

singku:
*Issue #, if available:*
`using StringView = std::string_view` was not taken into effect for C++ 17 
because the code is nested under another  "__cplusplus < 201703L"

*Description of changes:*

### DIFF
--- a/include/aws/crt/StringView.h
+++ b/include/aws/crt/StringView.h
@@ -7,21 +7,26 @@
 /**
  * string_view is introduced in C++ 17.
  */
-#if __cplusplus < 201703L
+#if __cplusplus >= 201703L
+#    include <string_view>
+namespace Aws
+{
+    namespace Crt
+    {
+        using StringView = std::string_view;
+    }
+} // namespace Aws
+#else
+
 #    include <algorithm>
 #    include <cassert>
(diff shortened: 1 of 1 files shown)


include/aws/crt/StringView.h
```diff
@@ -7,19 +7,21 @@
 /**
  * string_view is introduced in C++ 17.
  */
-#if __cplusplus >= 201703L
-#    include <string_view>
-#else
+#if __cplusplus < 201703L
```
graebm, line 10: trivial/debatable: simpler to just do 1 ifdef instead of 3?
```
#ifdef __cplusplus >= 201703L
#include <string_view>
namespace Aws
{
    namespace Crt
    {
        using StringView = std::string_view;
    }
}
#else
...rest of file...
#endif
```

graebm: I'd made this comment on the original PR.
I went back to find a link to it, so I could say "I told you so".
I discovered that I'd never actually submitted the comment.
🤦‍♂️🤦‍♂️🤦‍♂️

graebm (approved):
take my advice, or not

graebm (approved):

PR #175, opened 2020-09-17
Option to check headers at compile time.

graebm:
This is adapted from the header-checker in aws-c-common, but with separate passes for c++ 11,14,17,20.


TingDaoK (approved):

CMakeLists.txt
```diff
@@ -247,7 +255,6 @@ install(FILES ${AWS_CRT_IO_HEADERS} DESTINATION "include/aws/crt/io" COMPONENT D
 install(FILES ${AWS_CRT_IOT_HEADERS} DESTINATION "include/aws/iot" COMPONENT Development)
 install(FILES ${AWS_CRT_MQTT_HEADERS} DESTINATION "include/aws/crt/mqtt" COMPONENT Development)
 install(FILES ${AWS_CRT_HTTP_HEADERS} DESTINATION "include/aws/crt/http" COMPONENT Development)
-install(FILES ${AWS_CRT_EXTERNAL_HEADERS} DESTINATION "include/aws/crt/external" COMPONENT Development)
```
singku, line 250: I think it's worth to keep external headers in case customers want to use the same xml/json parsers we used so they don't have to do additional work like create their own external headers or do type conversion.

graebm: We're already hiding these symbols when compiling, so users shouldn't be able to use CJSON though these headers:

`target_compile_definitions(${PROJECT_NAME} PRIVATE -DCJSON_HIDE_SYMBOLS)`

But also, we **shouldn't** be offering CJSON to customers. It's not an official part of our API. We don't want to be obligated to keep them here forever. Like, even if we started offering the ION API in the future, we couldn't remove CJSON because we used to offer it as a feature.

PR #178, opened 2020-09-23
Support string_view

singku:
*Issue #, if available:*

*Description of changes:*


tests/StringViewTest.cpp
```diff
@@ -0,0 +1,127 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include <aws/crt/Api.h>
+#include <aws/crt/StringView.h>
+#include <aws/crt/Types.h>
+#include <aws/testing/aws_test_harness.h>
+
+static int s_test_string_view(struct aws_allocator *allocator, void *ctx)
```
justinboswell, line 10: Should we a) break these into separate tests, and b) test conversion to/from std::string_view and friends?

justinboswell (commented):
LGTM, just some small stuff and I'm ready to ship it

tests/StringViewTest.cpp
singku: Yeah, you are write, we should support conversion from std::string_view

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,845 @@
...
+} // namespace Aws
+
+// hash
+namespace std
+{
+    template <class CharT, class Traits>
+    struct hash<Aws::Crt::basic_string_view<CharT, Traits>>
+#if !defined(_MSC_VER)
+        : public unary_function<Aws::Crt::basic_string_view<CharT, Traits>, size_t>
+#endif
```
graebm, line 833: cppreference.com says [unary_function](https://en.cppreference.com/w/cpp/utility/functional/unary_function) is deprecated in c++11 and removed in c++17. Is it necessary to inherit from this for any compiler?

graebm (approved):

tests/StringViewTest.cpp
graebm: When we do this, we should probably check for std::string_view support like this:
```
#if __cplusplus >= 201703L || _MSC_LANG >= 201703L
```

As explained on [this page](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/) MSVC will only set the proper value for `__cplusplus` if the user adds the advanced `/Zc:__cplusplus` option to their program.

include/aws/crt/StringView.h
singku: Probably not if it's deprecated. I made it aligned for gcc and clang. But MSVC2019 is already not using it.

include/aws/crt/StringView.h
```diff
@@ -0,0 +1,845 @@
...
+
+            constexpr bool ends_with(CharT c) const noexcept
+            {
+                return !this->empty() && traits_type::eq(this->back(), c);
+            }
+
+            constexpr bool ends_with(const CharT *s) const noexcept { return this->ends_with(basic_string_view(s)); }
+
+            // find utilities
+            constexpr size_type find(const basic_string_view &s, size_type pos = 0) const noexcept
```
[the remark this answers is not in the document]
singku: It's not enforced but an indicator as do it whenever possible.

graebm (approved):

justinboswell (approved):

Issue #177, opened 2020-09-23
Crash in TlsContext::NewConnectionOptions() due to dereferencing a null pointer


EeroKurimo:
We've observed in some cases the underlying ``std::shared_ptr<aws_tls_ctx> m_ctx`` member of ``TlsContext`` may be storing a ``null`` when ``TlsContext::NewConnectionOptions()`` is called by ``MqttClient::NewConnection()``, which will cause the application crash. If in https://github.com/awslabs/aws-crt-cpp/blob/bd28bc4075d936c6fc49e60a8438dd25c1decead/source/io/TlsOptions.cpp#L286 the member variable ``m_ctx`` is storing null pointer, accessing ``m_ctx->alloc`` will crash in most compilers.

rccarper:
Hello, sorry that you're running into this.  I see a possible way this might happen, and am still investigating the best way to prevent a resulting crash.  Is it possible that you are sometimes passing a TlsContext that's not initialized to MqttClient::NewConnection?  (ie: the TlsContext was either created using the constructor with zero arguments, or copied from another TlsContext that was created in that way.)  Also, would it be possible to provide a callstack for the crash?  A log may be helpful as well.

EeroKurimo:
I found this while investigating another issue I opened in https://github.com/aws/aws-iot-device-sdk-cpp-v2/issues/162. Unfortunately I don't have access to work environment now so I can't get you a full backtrace until tomorrow, but in the one repro I saw this we were calling ``NewConnection(..., Io::TlsContext(tls_ctx_opt, Io::TlsMode::CLIENT))`` of https://github.com/awslabs/aws-crt-cpp/blob/f94efb11079eec6cf234d17430eafdb291930201/source/mqtt/MqttClient.cpp#L690 but the ``TlsContext`` ends up failing to initializing ``m_ctx`` when it tries to import the private key into the system in https://github.com/awslabs/aws-c-io/blob/2862e9ca07f5475b5490e645aec37ead8120ad96/source/darwin/darwin_pki_utils.c#L71.

[...]

PR #180, opened 2020-09-24
Better handling of uninitialized TlsContexts

rccarper:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-cpp/issues/177

*Description of changes:*
* Calling NewConnectionOptions on a TlsContext that's invalid (null context and/or initialization error) will now result in an error message and TlsConnectionOptions that are "uninitialized", ie, created with the no-arg constructor.
  * Previously, this was always assuming the context was non-null, which can cause a crash.
  * I went back-and-forth on using an assert instead of the above.  I don't necessarily disagree with that; I do like how that would be louder.  I ultimately went this way because I think it's awkward for us to make the no-arg constructor available (which sets it as "uninitialized" and with a null context) and guarantee a crash when someone calls NewConnectionOptions.  (Although we do certainly have similar behavior on TlsContextOptions.)
* Passing in an invalid TlsContext to MqttClient::NewConnection will now result in a null connection being returned.
  * I think that this makes sense, given that we provide an overload for NewConnection that does not require a TlsContext.
  * I was initially concerned that this might technically introduce a new crash for people currently passing in invalid TlsContexts, but those people should already be crashing when this flow eventually calls NewConnectionOptions.
  * I did look at making similar changes to the HTTP Connection/Manager code.  However, those take in an already built TlsConnectionOptions, which are allowed to be uninitialized (ie: uninitialized=don't use TLS).


source/io/TlsOptions.cpp
```diff
@@ -283,6 +284,13 @@ namespace Aws
 
             TlsConnectionOptions TlsContext::NewConnectionOptions() const noexcept
             {
+                if (!*this)
```
justinboswell, line 287: This gives me...feelings, but I get why it's here. Syntactically, it's a bit... obtuse. Is there not an IsValid() or similar we could call instead?

rccarper: Negative, the class just overloads the operator right now.  I can certainly add one though!

justinboswell (approved):

graebm (commented):
I'm not against this change, it's just that every function in this library has this issue 😕.
- Every member function could check that it's running on a valid instance.
- Every function that takes arguments could check that they are valid instances.

We chose not to use exceptions, so this is the world we live in.

source/mqtt/MqttClient.cpp
```diff
@@ -694,6 +694,15 @@ namespace Aws
                 const Crt::Io::TlsContext &tlsContext,
                 bool useWebsocket) noexcept
             {
+                if (!tlsContext)
+                {
+                    AWS_LOGF_ERROR(
+                        AWS_LS_MQTT_CLIENT,
+                        "id=%p Trying to call MqttClient::NewConnection using an invalid TlsContext.",
+                        (void *)m_client);
+                    return nullptr;
```
graebm, line 703: you also need to `aws_raise_error(AWS_ERROR_INVALID_ARGUMENT)`

rccarper: Added!

rccarper:
> I'm not against this change, it's just that every function in this library has this issue 😕.
> 
> * Every member function could check that it's running on a valid instance.
> * Every function that takes arguments could check that they are valid instances.
> 
> We chose not to use exceptions, so this is the world we live in.

I see where you're coming from.  I think minimally, MqttClient::NewConnection should still be validating its inputs.  I feel less strongly about the check in NewConnectionOptions, if we all feel that it's inconsistent/unnecessary.  But I do think we should at least be asserting in these cases.

graebm (approved):

PR #182, opened 2020-09-25
Imds

singku:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/aws/crt/ImdsClient.h
+++ b/include/aws/crt/ImdsClient.h
@@ -49,7 +49,7 @@ namespace Aws
             /**
              * https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
              */
-            struct AWS_CRT_CPP_API IamProfile
+            struct AWS_CRT_CPP_API IamProfileView
             {
                 DateTime lastUpdated;
                 StringView instanceProfileArn;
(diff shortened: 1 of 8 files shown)


tests/CMakeLists.txt
```diff
@@ -54,6 +54,14 @@ add_test_case(UUIDToString)
 add_test_case(TestIntArrayListToVector)
 add_test_case(TestByteCursorArrayListToVector)
 add_test_case(StringViewTest)
+add_test_case(TestIntArrayListToVector)
+add_test_case(TestByteCursorArrayListToVector)
```
graebm, line 58: `TestIntArrayListToVector` and `TestByteCursorArrayListToVector` are listed twice

graebm (approved):
fix & ship

justinboswell (approved):

include/aws/crt/ImdsClient.h
```diff
@@ -0,0 +1,441 @@
...
+                Vector<StringView> billingProducts;
+                StringView instanceType;
+                StringView accountId;
+                StringView imageId;
+                DateTime pendingTime;
+                StringView architecture;
+                StringView kernelId;
+                StringView ramdiskId;
+                StringView region;
+            };
```
graebm, line 108: Whoa, that's a lot of non-owning members!

Maybe we should rename this `InstanceInfoView` to hint that it doesn't own its memory?

Also, if the user wanted to persist this object, they'd need to declare their own type with 10+ String members and copy all the values over. We could provide that classe for them?

Or we could just have this use `String` for its members instead of `StringView`. It's not as efficient, but unless we expect this to be used in a very high-performance way, simplicity might win out.

singku: addressed this issue

include/aws/crt/ImdsClient.h
```diff
@@ -0,0 +1,441 @@
...
+
+            /**
+             * A convenient class for you to persist data from InstanceInfoView, which has StringView members.
+             */
+            struct AWS_CRT_CPP_API InstanceInfo
+            {
+                InstanceInfo() {}
+                InstanceInfo(const InstanceInfoView &other)
+                    : availabilityZone(other.availabilityZone.data(), other.availabilityZone.size()),
```
graebm, line 117: probably best to move these function implementations to the cpp file, to keep down the cost of including this header

graebm (approved):
fix & ship

PR #183, opened 2020-09-29
Moved aws-common-runtime -> crt

justinboswell:


singku:
What's the story behind this move?

justinboswell:
> What's the story behind this move?

shortening submodule paths, which will shorten the recursive dirs that cmake creates, which will allow windows users to have slightly more permissive paths to keep under the 260 MAX_PATH length.

.gitmodules
```diff
@@ -1,30 +1,30 @@
 [submodule "aws-common-runtime/aws-c-common"]
```
singku, line 1: Didn't change this for readability?

singku (approved):

.gitmodules
justinboswell: Didn't change it because it's just a name, and it would break the history.

PR #187, opened 2020-10-01
Fix symbol import/export issues on Windows.

wps132230:
**Issue 1**:
> error C2491: 'Aws::Crt::ArrayListToVector': definition of dllimport function not allowed
```c++
// In include/aws/crt/Types.h
template <typename RawType, typename TargetType>
AWS_CRT_CPP_API Vector<TargetType> ArrayListToVector(const aws_array_list *array, TypeConvertor<RawType, TargetType> conv)
```
**Fix issue 1**:
Remove `AWS_CRT_CPP_API`

**Issue 2**:
> warning C4273: 'Aws::Crt::TestCleanupAndWait': inconsistent dll linkage
> warning LNK4217: locally defined symbol TestCleanupAndWait
```c++
// In both include/aws/crt/Api.h and tests/Utils.h
AWS_CRT_CPP_API void TestCleanupAndWait();
```
**Fix issue 2**:
1. Remove `TestCleanupAndWait` in `include/aws/crt/Api.h`
2. Remove `AWS_CRT_CPP_API` for `TestCleanupAndWait` in `tests/Utils.h`
3. `#include <Utils.h>` for all tests.

**Issue 3**:
>  Policy CMP0077 is not set: option() honors normal variables.

This warning happens when I use `aws-crt-cpp` as a submodule of `aws-sdk-cpp`. Starting from `CMake` version 3.13, If I set a normal variable `BUILD_TESTING` in `aws-sdk-cpp`'s `CMakeLists.txt`. This variable will be removed if `aws-crt-cpp`'s `CMakeLists.txt` has `include(CTest)` because `BUILD_TESTING` is a cache entry in `option`: https://github.com/Kitware/CMake/blob/ad478a4a398d6048c2cb9dbb05c1aecd8c31608e/Modules/CTest.cmake#L50
To summary, the issue is we set `BUILD_TESTING` as `OFF` in `aws-sdk-cpp`, but it's still `ON` in `aws-crt-cpp`.

**Fix issue 3**:
https://cmake.org/cmake/help/git-stage/policy/CMP0077.html
Set CMP0077 to `NEW`, then it will do nothing with `option(BUILD_TESTING)`. The normal variable `BUILD_TESTING` will not be changed.

### DIFF
--- a/include/aws/crt/Types.h
+++ b/include/aws/crt/Types.h
@@ -77,9 +77,7 @@ namespace Aws
          * A conversion function should be provided to do the type conversion
          */
         template <typename RawType, typename TargetType>
-        Vector<TargetType> ArrayListToVector(
-            const aws_array_list *array,
-            TypeConvertor<RawType, TargetType> conv)
+        Vector<TargetType> ArrayListToVector(const aws_array_list *array, TypeConvertor<RawType, TargetType> conv)
         {
             Vector<TargetType> v;
             size_t cnt = aws_array_list_length(array);
--- a/include/aws/crt/Types.h
+++ b/include/aws/crt/Types.h
@@ -77,7 +77,7 @@ namespace Aws
          * A conversion function should be provided to do the type conversion
          */
         template <typename RawType, typename TargetType>
-        AWS_CRT_CPP_API Vector<TargetType> ArrayListToVector(
+        Vector<TargetType> ArrayListToVector(
             const aws_array_list *array,
             TypeConvertor<RawType, TargetType> conv)
         {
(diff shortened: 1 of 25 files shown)


bretambrose (approved):

include/aws/crt/Types.h
```diff
@@ -77,9 +77,7 @@ namespace Aws
          * A conversion function should be provided to do the type conversion
          */
         template <typename RawType, typename TargetType>
-        AWS_CRT_CPP_API Vector<TargetType> ArrayListToVector(
-            const aws_array_list *array,
-            TypeConvertor<RawType, TargetType> conv)
+        Vector<TargetType> ArrayListToVector(const aws_array_list *array, TypeConvertor<RawType, TargetType> conv)
```
singku, line 80: This removal makes much sense. It's implemented in header file, no need to export.

singku (approved):

PR #193, opened 2020-10-14
Add aws-c-iot dependency and Device Defender bindings

giordzaws:
*Description of changes:*

* Added [aws-c-iot](https://github.com/awslabs/aws-c-iot) package as a dependency
* Added bindings for Device Defender metric reporting, with core functionality implemented in the aws-c-iot package
* Added simple tests for the Device Defender bindings, mostly around ensuring correct state of the task
* Removed the android-armv7 target from CI since it's building for an older API level 19 Android KitKat


.gitmodules
```diff
@@ -28,3 +28,6 @@
 [submodule "aws-common-runtime/aws-c-event-stream"]
 	path = crt/aws-c-event-stream
 	url = https://github.com/awslabs/aws-c-event-stream.git
+[submodule "aws-common-runtime/aws-c-iot"]
+	path = crt/aws-c-iot
+	url = https://github.com/awslabs/aws-c-iot.git
```
giordzaws, line 33: Will need to update the aws-c-iot commit once the changes from my remote branch in aws-c-iot are merged into main.

include/aws/crt/iot/DeviceDefender.h
```diff
@@ -0,0 +1,200 @@
...
+
+              private:
+                std::shared_ptr<Mqtt::MqttConnection> mqttConnection;
+                ByteCursor thingName;
+                Io::EventLoopGroup &eventLoopGroup;
+                DeviceDefenderReportFormat reportFormat;
+                uint64_t taskPeriodNs;
+                uint64_t networkConnectionSamplePeriodNs;
+                OnDefenderV1TaskCancelledHandler onCancelled;
+                void *cancellationUserdata;
```
fufranci, line 84: Consider to reuse existing C `struct aws_iotdevice_defender_report_task_config` **if possible**.

include/aws/crt/iot/DeviceDefender.h
```diff
@@ -0,0 +1,200 @@
...
+            {
+              public:
+                DeviceDefenderV1ReportTask(
+                    Aws::Crt::Allocator *allocator,
+                    const DeviceDefenderV1ReportTaskConfig &config) noexcept;
+                ~DeviceDefenderV1ReportTask();
+                DeviceDefenderV1ReportTask(const DeviceDefenderV1ReportTask &) = delete;
+                DeviceDefenderV1ReportTask(DeviceDefenderV1ReportTask &&) = delete;
+                DeviceDefenderV1ReportTask &operator=(const DeviceDefenderV1ReportTask &) = delete;
+                DeviceDefenderV1ReportTask &operator=(DeviceDefenderV1ReportTask &&) = delete;
```
fufranci, line 164: Is this the move assignment operator? Would it offer value if we keep this? It am guessing the implementation is simple.

include/aws/crt/iot/DeviceDefender.h
giordzaws: That makes sense. I've refactored to reuse the existing struct in updated commits.

include/aws/crt/iot/DeviceDefender.h
giordzaws: Good point, it could be of use. I've added implementation for the move operator in updated commits.

include/aws/crt/auth/Credentials.h
```diff
@@ -125,7 +125,12 @@ namespace Aws
...
-                CredentialsProviderStaticConfig() : AccessKeyId{}, SecretAccessKey{}, SessionToken{} {}
+                CredentialsProviderStaticConfig()
+                {
+                    AWS_ZERO_STRUCT(AccessKeyId);
+                    AWS_ZERO_STRUCT(SecretAccessKey);
+                    AWS_ZERO_STRUCT(SessionToken);
+                }
 
                 /**
                  * The value of the access key component for the provider's static aws credentials
```
giordzaws, line 136: This was somehow pulled in from merged commit: https://github.com/awslabs/aws-crt-cpp/commit/8e4da9be463cd74d1a9f9f7ff4f113f3b85c4c9a

source/io/Stream.cpp
```diff
@@ -114,7 +114,7 @@ namespace Aws
                 auto read = m_stream->gcount();
                 buffer.len += static_cast<size_t>(read);
 
-                if (read > 0)
+                if (read > 0 || (read == 0 && m_stream->eof()))
```
giordzaws, line 117: This was somehow pulled in from merged commit: https://github.com/awslabs/aws-crt-cpp/commit/8e4da9be463cd74d1a9f9f7ff4f113f3b85c4c9a

tests/StreamTest.cpp
```diff
@@ -87,6 +87,34 @@ static int s_StreamTestRead(struct aws_allocator *allocator, void *ctx)
...
+        aws_byte_buf_clean_up(&buffer);
+    }
+
+    Aws::Crt::TestCleanupAndWait();
+
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(StreamTestReadEmpty, s_StreamTestReadEmpty)
+
```
giordzaws, line 117: This was somehow pulled in from merged commit: https://github.com/awslabs/aws-crt-cpp/commit/8e4da9be463cd74d1a9f9f7ff4f113f3b85c4c9a

source/iot/DeviceDefender.cpp
```diff
@@ -0,0 +1,210 @@
...
+            {
+                if (this->GetStatus() == DeviceDefenderV1ReportTaskStatus::Ready)
+                {
+
+                    this->m_owningTask = aws_iotdevice_defender_v1_report_task(this->m_allocator, &this->m_taskConfig);
+
+                    if (this->m_owningTask == nullptr)
+                    {
+                        this->m_lastError = aws_last_error();
+                        this->m_status = DeviceDefenderV1ReportTaskStatus::Failed;
```
giordzaws, line 117: Question for reviewers: Is setting a failed status for the task and triggering the callback preferable to throwing en exception?

include/aws/crt/iot/DeviceDefender.h
```diff
@@ -0,0 +1,184 @@
...
+            {
+
+                friend class DeviceDefenderV1ReportTask;
+                friend class DeviceDefenderV1ReportTaskConfigBuilder;
+
+              public:
+                /**
+                 * Creates a client configuration for use with making new AWS Iot specific MQTT Connections with MTLS.
+                 */
+                DeviceDefenderV1ReportTaskConfig(
```
fufranci, line 61: This could be private since there is a builder, unless we expect customer to use this constructor directly.

include/aws/crt/iot/DeviceDefender.h
```diff
@@ -0,0 +1,184 @@
...
+              private:
+                OnDefenderV1TaskCancelledHandler onCancelled;
+                void *cancellationUserdata;
+                aws_iotdevice_defender_report_task_config taskConfig;
+            };
+
+            /**
+             * Represents a builder for creating a DeviceDefenderV1ReportTaskConfig object.
+             */
+            class AWS_CRT_CPP_API DeviceDefenderV1ReportTaskConfigBuilder final
```
fufranci, line 80: I am wondering if this makes sense:

1. Remove DeviceDefenderV1ReportTaskConfig and DeviceDefenderV1ReportTaskConfigBuilder.
1. Create DeviceDefenderV1ReportTaskBuilder
1. The TaskBuilder interface would look similar with ConfigBuilder.
1. The usage would be something like:
```
TaskBuilder taskBuilder;
Task t = taskBuilder.withA().withB().Build();
```

So the idea is to remove the exposure of config altogether.

fufranci: BTW, I have **NOT** checked what is the existing pattern for the rest of the CRT.

giordzaws: Yeah I think that could make sense. I've thought about that pattern as well and hiding the config within the Task class so that Task is the only exposed class. The builder pattern could still ensure that the Task is given all required parameters, and that might clean things up. I'll give that a try with a follow up commit. Thanks!

include/aws/crt/iot/DeviceDefender.h
giordzaws: Good point, with the Builder class refactored to create an actual DD Task instead of just a Config class, I think it makes sense to make the DDTask constructor private so only the Builder calls it.

include/aws/crt/mqtt/MqttClient.h
```diff
@@ -107,6 +113,7 @@ namespace Aws
             class AWS_CRT_CPP_API MqttConnection final
             {
                 friend class MqttClient;
+                friend class Iot::DeviceDefenderV1ReportTask;
```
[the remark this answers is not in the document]
DavidOgunsAWS: I see why this is needed now. I personally think the exposure of the underlying mqtt_client pointer through a member function is somewhat better but still flawed. I'll consider this OK

include/aws/crt/iot/DeviceDefender.h
```diff
@@ -0,0 +1,169 @@
...
+             * Represents a builder for creating a DeviceDefenderV1ReportTask object.
+             */
+            class AWS_CRT_CPP_API DeviceDefenderV1ReportTaskBuilder final
+            {
+              public:
+                DeviceDefenderV1ReportTaskBuilder(
+                    Aws::Crt::Allocator *allocator,
+                    std::shared_ptr<Mqtt::MqttConnection> mqttConnection,
+                    Io::EventLoopGroup &eventLoopGroup,
+                    ByteCursor thingName);
```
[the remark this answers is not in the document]
giordzaws: Good point, I see other classes such as the iot MqttClient class that accept a Crt::String parameter as part of the exposed APIs. I'll update this code to use a Crt::String and do the conversion internally.

source/iot/DeviceDefender.cpp
giordzaws: Ok, that makes sense. It looks like we expect users to [check for return codes](https://github.com/aws/aws-iot-device-sdk-cpp-v2/blob/master/samples/mqtt/basic_pub_sub/main.cpp#L471) on methods to see if the method returned with an error code. I'll remove the Failed state and update this to raise an error.

source/iot/DeviceDefender.cpp
```diff
@@ -0,0 +1,197 @@
...
+                ByteCursor thingName,
+                Io::EventLoopGroup &eventLoopGroup,
+                DeviceDefenderReportFormat reportFormat,
+                uint64_t taskPeriodNs,
+                uint64_t networkConnectionSamplePeriodNs,
+                OnDefenderV1TaskCancelledHandler &&onCancelled,
+                void *cancellationUserdata) noexcept
+                : OnDefenderV1TaskCancelled(std::move(onCancelled)), cancellationUserdata(cancellationUserdata),
+                  m_allocator(allocator), m_status(DeviceDefenderV1ReportTaskStatus::Ready),
+                  m_taskConfig{mqttConnection.get()->m_underlyingConnection,
```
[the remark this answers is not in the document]
giordzaws: I believe this is resolved with the conversation regarding the friend class in the MqttClient. Let me know if that's not the case.

DavidOgunsAWS (approved):

include/aws/crt/iot/DeviceDefender.h
```diff
@@ -0,0 +1,168 @@
...
+        namespace Iot
+        {
+
+            class DeviceDefenderV1ReportTask;
+            class DeviceDefenderV1ReportTaskBuilder;
+
+            /**
+             * Invoked upon DeviceDefender V1 task cancellation.
+             */
+            using OnDefenderV1TaskCancelledHandler = std::function<void(void *)>;
```
justinboswell, line 35: Do we want to shove the V1 stuff in a V1 namespace and not pollute the names?

include/aws/crt/iot/DeviceDefender.h
```diff
@@ -0,0 +1,168 @@
...
+
+                /**
+                 * Sets the task period nanoseconds. Defaults to 5 minutes.
+                 */
+                DeviceDefenderV1ReportTaskBuilder &WithTaskPeriodNs(uint64_t taskPeriodNs) noexcept;
+
+                /**
+                 * Sets the network connection sample period nanoseconds. Defaults to 5 minutes.
+                 */
+                DeviceDefenderV1ReportTaskBuilder &WithNetworkConnectionSamplePeriodNs(
```
justinboswell, line 134: Why is nano-seconds the sample period unit? Why not something closer to our actual expected max sample rate, say...seconds? Maybe milliseconds, but that would be a lot of sampling.

tests/DeviceDefenderTest.cpp
```diff
@@ -0,0 +1,141 @@
...
+            .WithDefenderV1TaskCancellationUserData(&data);
+
+        Aws::Crt::Iot::DeviceDefenderV1ReportTask task = taskBuilder.Build();
+
+        ASSERT_INT_EQUALS((int)Aws::Crt::Iot::DeviceDefenderV1ReportTaskStatus::Ready, (int)task.GetStatus());
+
+        task.StartTask();
+
+        ASSERT_INT_EQUALS((int)Aws::Crt::Iot::DeviceDefenderV1ReportTaskStatus::Running, (int)task.GetStatus());
+        std::this_thread::sleep_for(std::chrono::seconds(1));
```
justinboswell, line 62: these sleeps are guaranteed to eventually fail in CI (sleep is really variadic on virtualized hosts), and networks are...stochastic. Use condition_variables or another signaling mechanism to continue execution when the desired state is reached.

justinboswell (changes requested):

include/aws/crt/iot/DeviceDefender.h
giordzaws: Good point, I'll create a DeviceDefenderV1 namespace and move things under that.

include/aws/crt/iot/DeviceDefender.h
giordzaws: That makes sense. I think seconds as the unit is appropriate. I'll make that change and have it do the conversion to nanos for passing to the C layer.

tests/DeviceDefenderTest.cpp
giordzaws: Ok, I'll modify this to use a condition variable in a follow up commit.

justinboswell (approved):

xiazhvera:
The changes mentioned in this PR has been brought into IoT cpp v2 SDK: https://github.com/aws/aws-iot-device-sdk-cpp-v2/pull/184

PR #195, opened 2020-10-27
Wire APIs and callbacks to the C layer

fufranci:
These changes should go to aws-iot-device-sdk-cpp-v2. I am using this PR to test my changes in the meantime.

### DIFF
--- a/include/aws/crt/iot/SecureTunnel.h
+++ b/include/aws/crt/iot/SecureTunnel.h
@@ -27,7 +27,6 @@ namespace Aws
               public:
                 SecureTunnel(
                     // Parameters map to aws_secure_tunneling_connection_config
-
                     Allocator *allocator,
                     Aws::Crt::Io::ClientBootstrap *clientBootstrap,
                     Aws::Crt::Io::SocketOptions *socketOptions,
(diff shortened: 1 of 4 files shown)


include/aws/crt/iot/SecureTunnel.h
```diff
@@ -0,0 +1,79 @@
...
+    namespace Crt
+    {
+        namespace Iot
+        {
+            // Callbacks
+            using OnSecureTunnelingConnectionComplete = std::function<void(void)>;
+            using OnSecureTunnelingSendDataComplete = std::function<void(int errorCode)>;
+            using OnSecureTunnelingDataReceive = std::function<void(const ByteBuf &data)>;
+            using OnSecureTunnelingStreamStart = std::function<void(int32_t streamId)>;
+            using OnSecureTunnelingStreamReset = std::function<void(void)>;
```
DavidOgunsAWS, line 22: Depending on the API's intent on what to manage, either stream ID's should be on all of these callbacks, or none of them. If the underlying handling is tracking the active valid stream ID, and properly ignores all messages of a non-matching stream IDs*, then above it stream ID has no value to pass along since it will only ever see valid data, and reset messages.

*StreamStart which indicates there is a new valid StreamID, implicitly closing the previous one

PR #213, opened 2021-02-05
Add retain, dup, qos to MQTT on-message callback.

graebm:
This is a backwards compatible change.

Mark existing `OnPublishReceivedHandler` deprecated, replace with `OnMessageReceivedHandler`. New handler carries 3 more args: `retain`, `qos`, and `dup`. Mark functions that take old handler as deprecated, and add otherwise identical function that accept new handler.

### DIFF
--- a/source/mqtt/MqttClient.cpp
+++ b/source/mqtt/MqttClient.cpp
@@ -66,7 +66,7 @@ namespace Aws
                 PubCallbackData() : connection(nullptr), allocator(nullptr) {}
 
                 MqttConnection *connection;
-                OnPublishReceivedHandler onPublishReceived;
+                OnMessageReceivedHandler onPublishReceived;
                 Allocator *allocator;
             };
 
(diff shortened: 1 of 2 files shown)


sdavtaker (changes requested):
Added a few variablel renaming comments for readability.

source/mqtt/MqttClient.cpp
```diff
@@ -427,6 +431,15 @@ namespace Aws
             }
 
             bool MqttConnection::SetOnMessageHandler(OnPublishReceivedHandler &&onPublish) noexcept
+            {
+                return SetOnMessageHandler(
+                    [onPublish](
```
bretambrose, line 436: Is the capture here correct?  Memory is fuzzy wrt capture movable type.

source/mqtt/MqttClient.cpp
```diff
@@ -66,7 +66,7 @@ namespace Aws
                 PubCallbackData() : connection(nullptr), allocator(nullptr) {}
 
                 MqttConnection *connection;
-                OnPublishReceivedHandler onPublishReceived;
+                OnMessageReceivedHandler onPublishReceived;
```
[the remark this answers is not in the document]
graebm: Good call. Didn't do it before just to keep the diff as simple as possible.
But yeah, makes the code more understandable going forward from here

source/mqtt/MqttClient.cpp
graebm: Yeah, it just makes a copy of onPublish.

If we were using c++14, we could do a std::move when capturing, but alas

sdavtaker (approved):

PR #229, opened 2021-03-24
Adding protection around return expression

DavidOgunsAWS:
*Issue #, if available:*
[This line](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/StlAllocator.h#L50) has a signed/unsigned conversion issue.

*Description of changes:*
My suspicion for the find is that the code base in question had defined std::allocator<T>::pointer for their own type and it was a signed integral type. There may be platform specific reasons to do this, however our code needs to be aware and behave appropriate. I don't believe static_cast<> from void * to a signed integer type is the proper fix, but we may want to static_assert against doing this until there is one, or we can implement the proper measures around the check.


include/aws/crt/StlAllocator.h
```diff
@@ -47,7 +48,11 @@ namespace Aws
             {
                 (void)hint;
                 AWS_ASSERT(m_allocator);
-                return reinterpret_cast<RawPointer>(aws_mem_acquire(m_allocator, n * sizeof(T)));
+                if (reinterpret_is_safe())
+                {
+                    return reinterpret_cast<RawPointer>(aws_mem_acquire(m_allocator, n * sizeof(T)));
+                }
+                return static_cast<RawPointer>(aws_mem_acquire(m_allocator, n * sizeof(T)));
```
graebm, line 55: would it be safe to just ALWAYS do static_cast?
wouldn't that automagically compile to be equivalent to reinterpret_cast, if it's safe to do reinterpret_cast?

DavidOgunsAWS: Static cast seems to be the way to go!

justinboswell (approved):

PR #231, opened 2021-04-01
BYO_CRYPTO

graebm:
Allow users to supply their own TLS and crypto providers, instead of using the defaults ([s2n-tls](https://github.com/aws/s2n-tls) on Linux, [Secure Transport](https://developer.apple.com/documentation/security/secure_transport) on Apple, [Secure Channel](https://docs.microsoft.com/en-us/windows/win32/secauthn/secure-channel) on Windows).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #232, opened 2021-04-01
Misc fixes

graebm:
- additional CI runs: SHARED_LIBS=ON, Visual Studio versions
- Aws::Crt::Optional cannot alias std::optional. This library is always built with c++11, so always uses our definition of Optional. If a user's c++17 app includes this file, they'll think that instances created by our library are ACTUALLY std::optional, but these have different binary layout, so kaboom


graebm:
These changes to Optional ended up being incorporated by this PR: https://github.com/awslabs/aws-crt-cpp/pull/231

The shared-libs CI changed were incorporated by this PR: https://github.com/awslabs/aws-crt-cpp/pull/299

PR #233, opened 2021-04-02
deep copy of the payload for user

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/mqtt/MqttClient.cpp
+++ b/source/mqtt/MqttClient.cpp
@@ -130,6 +130,7 @@ namespace Aws
                 if (callbackData->payload)
                 {
                     aws_byte_buf_clean_up(callbackData->payload);
+                    Crt::Delete(callbackData->payload, callbackData->allocator);
                 }
 
                 Crt::Delete(callbackData, callbackData->allocator);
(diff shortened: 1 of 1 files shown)


source/mqtt/MqttClient.cpp
```diff
@@ -104,6 +104,7 @@ namespace Aws
                 MqttConnection *connection;
                 OnOperationCompleteHandler onOperationComplete;
                 const char *topic;
+                ByteBuf *payload;
```
DavidOgunsAWS, line 107: This should not be a pointer type. Then it needs to be initialized to an empty/blank buffer `{ allocator = nullptr, len = 0, ptr = nullptr }` so a call to cleanup if it's still in that state is still a safe no-op.

source/mqtt/MqttClient.cpp
```diff
@@ -674,9 +682,12 @@ namespace Aws
                 opCompleteCallbackData->allocator = m_owningClient->allocator;
                 opCompleteCallbackData->onOperationComplete = std::move(onOpComplete);
                 opCompleteCallbackData->topic = topicCpy;
+                opCompleteCallbackData->payload = Crt::New<ByteBuf>(m_owningClient->allocator);
```
DavidOgunsAWS, line 685: redundant: assuming payload isn't a `ByteBuf *` then this goes away entirely. So does the call to check if it is null before calling cleanup. Just call cleanup.

justinboswell (approved):

PR #234, opened 2021-04-08
Improve in-source build for aws-lc.

wps132230:
*Issue:*
Existing in-source build will run CMake scripts via `execute_process` to build and install `aws-lc` during CMake process.
In theory, the configuration step should not build or install anything.
There are several issues need to be solved because of that:
1. CMake requires sudo permission if `CMAKE_INSTALL_PREFIX` is not specified, as it installs `aws-lc` into system directory.
2. CMake parameters in `aws-lc.cmake` are very limited, thus not working with some special platforms or cross compile, for example Android.

*Description of changes:*
See PRs: https://github.com/awslabs/aws-lc/pull/127, https://github.com/aws/s2n-tls/pull/2714 and https://github.com/awslabs/aws-c-cal/pull/82
1. Add `aws-lc` as subdirectory rather than executing commands in `aws-lc.cmake` to build and install `aws-lc`.
2. Turn on `DISABLE_GO`, which is required for Android build for `aws-lc`.
3. Bring back Android build in CI.


justinboswell (approved):

wps132230:
Closing this PR, as it's a part of https://github.com/awslabs/aws-crt-cpp/pull/243 now.

PR #236, opened 2021-04-15
Enable MQTT client to use TPM certificates in Windows

DavidOgunsAWS:
Add the ability to use a client certificate located in a Windows certificate store. Previously, the client certificate and private key had to be passed by filepath or file contents. With this change, certificates and keys stored on TPM devices can be used.


include/aws/iot/MqttClient.h
```diff
@@ -143,6 +143,22 @@ namespace Aws
           public:
             MqttClientConnectionConfigBuilder();
 
+#   ifdef _WIN32
+            /**
+             * Sets the builder up for MTLS using system certificate path. These are certitficates managed by
+             * WindowsCertStoreMgr
+             * 
+             * Example system path value for an installed user certificate: 
+             *   "CurrentUser\My\6ac133ac58f0a88b83e9c794eba156a98da39b4c"
```
bretambrose, line 152: Is there something conceptually analagous in other platforms?  I'd really like to see us move from platform ifdefs in header files to empty implementations in non-matching platforms.  Just curious.

DavidOgunsAWS: Not yet. So this PR may not actually get pushed. I should probably keep this as a branch.

sbSteveK (approved):

Issue #204, opened 2020-12-15
Support custom keychain on macOS


EeroKurimo:
In aws/aws-iot-device-sdk-cpp-v2#163 I suggested adding support for using a custom keychain instead of the default login keychain for use-cases like daemons using the AWS IoT SDK. It was suggested that I'd instead change the default keychain for the account to the private keychain, or just a PKCS#12 file which appeared to not use a keychain.

As commented in the ticket, now when doing the daemon integration, we wound that changing the default keychain of the daemon account appeared to cause some other issues, and using PKCS#12 appears to also rely on the default keychain. To avoid dealing with the default keychain, I'd wish to add support for optionally using a custom private keychain for the IoT TLS connections only.

This issue covers the changes for aws-crt-cpp to support the changes in https://github.com/awslabs/aws-c-io/issues/340

EeroKurimo:
This pull request can be ignored until https://github.com/awslabs/aws-c-io/pull/341 is wrapped up, as any changes there may need to be reflected here.

bretambrose:
Previously merged.

PR #239, opened 2021-04-21
Custom keychain support for macOS TLS (#204)

EeroKurimo:
*Issue #204*

*Description of changes:*

An optional macos keychain path for macOS TLS connection was added in https://github.com/awslabs/aws-c-io/pull/368. This change provides the C++ method for changing that value for a `TlsContextOptions` object.

As an implementation note, I noticed that all the other `Set*` and `Override*` methods are using a `const ByteCursor &`, but this change uses a non-const `ByteCursor &` due to `aws_tls_ctx_options_set_keychain_path()` also using a non-const `struct aws_byte_cursor` (not a pointer!) as its argument. It actually looks like it should be `const struct aws_byte_cursor *` instead, as it's only copied with `aws_string_new_from_cursor()` that expects a `const struct aws_byte_cursor *`. I'm open to both merging this and fixing that later, or changing that first in aws-c-io before changing aws-crt-cpp.

### DIFF
--- a/include/aws/crt/io/TlsOptions.h
+++ b/include/aws/crt/io/TlsOptions.h
@@ -88,6 +88,14 @@ namespace Aws
                     const char *pkcs12_path,
                     const char *pkcs12_pwd,
                     Allocator *allocator = g_allocator) noexcept;
+
+                /**
+                 * By default the certificates and private keys are stored in the default keychain
+                 * of the account of the process. If you instead wish to provide your own keychain
+                 * for storing them, this makes the TlsContext to use that instead.
+                 * NOTE: The password of your keychain must be empty.
+                 */
+                bool SetKeychainPath(ByteCursor &keychain_path) noexcept;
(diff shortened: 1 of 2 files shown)


bretambrose:
At the time, the move to cursor-by-value was intentional; that's a pattern we've moved to since it's tedious and annoying to have to declare first and then pass the pointer.  On the other hand, that leads to an inconsistent tls configuration API.  So viewed now from a distance, I think I should switch it back to by-pointer for consistency's sake.  The usage improvement from by-value is not enough to justify breaking consistency.

I'll try to finish that up this morning.

EeroKurimo:
Thanks for the explanation! Passing the struct by value indeed would've opened the nice option of passing rvalues. However I also agree that having a consistent API makes things smoother for the end user, so sticking with pointers might be better for now.

bretambrose:
Merged in previous PR.

PR #247, opened 2021-05-13
Add support to store and pass SDK Name & Version as a metric to IoT c…

HarshGandhi-AWS:
…loud

*Issue #, if available:*

*Description of changes:*

With this change request, I am adding support to pass and store SDK Name and its Version so we can pass it to IoT Cloud while making mqtt connection. This will give ability to external software running CPPv2 to pass their own metric so we can effectively track them. 

### DIFF
--- a/source/iot/MqttClient.cpp
+++ b/source/iot/MqttClient.cpp
@@ -403,8 +403,10 @@ namespace Aws
                 return nullptr;
             }
 
-            char username[100] = "";
-            sprintf(username, "%s%s%s%s", "?SDK=", config.m_sdkName.c_str(), "&Version=", config.m_sdkVersion.c_str());
+            size_t username_size = sizeof(config.m_sdkName.c_str()) + sizeof(config.m_sdkVersion.c_str()) + 15;
+            char *username = new char[username_size];
+            snprintf(
+                username, username_size, "%s%s%s%s", "?SDK=", config.m_sdkName.c_str(), "&Version=", config.m_sdkVersion.c_str());
 
             if (!(*newConnection) || !newConnection->SetLogin(username, nullptr))
             {
@@ -421,6 +423,8 @@ namespace Aws
             {
                 newConnection->SetHttpProxyOptions(config.m_proxyOptions.value());
             }
+
+            delete[] username;
 
             return newConnection;
         }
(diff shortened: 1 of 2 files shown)


source/iot/MqttClient.cpp
```diff
@@ -352,7 +403,10 @@ namespace Aws
                 return nullptr;
             }
 
-            if (!(*newConnection) || !newConnection->SetLogin("?SDK=CPPv2&Version=" AWS_CRT_CPP_VERSION, nullptr))
+            char username[100] = "";
+            sprintf(username, "%s%s%s%s", "?SDK=", config.m_sdkName.c_str(), "&Version=", config.m_sdkVersion.c_str());
```
DavidOgunsAWS, line 407: The string length of username is probably something that should be looked up with MQTT spec maximum. Maybe less than spec if the spec is unreasonably high.

Use snprintf() instead of sprintf().

bretambrose (commented):
Why do we need new constructors for MqttClientConnectionConfig?  Shouldn't the builder, with its friended access, make that unnecessary?

HarshGandhi-AWS:
@bretambrose Added new constructors for backward compatibility. Not happy about the redundant code but didn't find any other workaround for it.

bretambrose:
> @bretambrose Added new constructors for backward compatibility. Not happy about the redundant code but didn't find any other workaround for it.



> @bretambrose Added new constructors for backward compatibility. Not happy about the redundant code but didn't find any other workaround for it.

The builder should be able to mutate/access members on a config.  It's not pretty, but one of the selling points of a builder is removing constructor explosion.

source/iot/MqttClient.cpp
HarshGandhi-AWS: Made changes as requested.

graebm (approved):
I just jumped in and made changes, since I have a bit of a big-picture view on this code, and similar code in the other languages.

I kept the specifics of the metrics string solely in the Builder class. I half-added the concept of username and password to the Config that the builder creates. I say "half-added" because it's not publicly exposed to users at the moment.

HarshGandhi-AWS:
@graebm Thank you for updating the code. I went ahead and tested the changes just to be sure. They are working fine and I think are good to merge in.

bretambrose (approved):

DavidOgunsAWS (approved):

PR #250, opened 2021-05-14
Added missing limits include

jbrownsw:
This was causing compilation errors on gcc 11.1 x86. I'm not sure why my x64 build was fine with it, but clearly the include should be there since numeric_limits is used.

*Issue #, if available:*

*Description of changes:*


bretambrose:
Merged in https://github.com/awslabs/aws-crt-cpp/pull/254

PR #255, opened 2021-05-26
stop using deprecated aws_off_t

graebm:
related:
https://github.com/awslabs/aws-c-common/pull/808
https://github.com/awslabs/aws-c-io/pull/386


include/aws/crt/io/Stream.h
```diff
@@ -15,6 +15,10 @@ namespace Aws
         namespace Io
         {
             using StreamStatus = aws_stream_status;
+
+            /**
+             * @deprecated Use int64_t instead for offsets in public APIs.
+             */
             using OffsetType = aws_off_t;
```
sdavtaker, line 22: Please mark it with __attribute__((deprecated("message", "replacement")));
or maybe just replace the type here in place of change it everywhere.

bretambrose (approved):

include/aws/crt/io/Stream.h
graebm: Good suggestion, but aws-crt-cpp supports c++11 for now, and it looks like they didn't get it right until c++14/17 😭

GCC/Clang only: `__attribute__(deprecated)`
MSVC only: `__declspec(deprecated)`
c++11: supports attributes`[[]]`, but not `deprecated`
c++14: supports `[[deprecated("reason")]]`
c++17: unknown attributes are ignored and not considered errors <-- they finally got it right. 6 years later.

I did remove all uses of `OffsetType` in aws-crt-cpp, but I'm leaving the declaration here so I don't break anything that was using it downstream

I could put together a macro [like this](https://stackoverflow.com/a/21265197), but I think I'll just wait for the tides of time to come and make our minspec c++14

PR #276, opened 2021-09-17
Making S3 client optional on the build

sdavtaker:
*Issue #, if available:*

*Description of changes:* Making S3 client optional on the build

### DIFF
--- a/source/Api.cpp
+++ b/source/Api.cpp
@@ -12,7 +12,9 @@
 #include <aws/event-stream/event_stream.h>
 #include <aws/http/http.h>
 #include <aws/mqtt/mqtt.h>
-#include <aws/s3/s3.h>
+#ifdef ENABLE_S3
+    #include <aws/s3/s3.h>
+#endif
 
 namespace Aws
 {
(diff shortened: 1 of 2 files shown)


source/Api.cpp
```diff
@@ -37,7 +39,9 @@ namespace Aws
             // sets up the StlAllocator for use.
             g_allocator = allocator;
             aws_mqtt_library_init(allocator);
-            aws_s3_library_init(allocator);
+            #ifdef ENABLE_S3_BUILT
+                aws_s3_library_init(allocator);
```
wps132230, line 43: `aws_s3_library_init` initializes several other libraries, including `aws_auth_library_init` and `aws_http_library_init`.

source/Api.cpp
```diff
@@ -71,7 +75,9 @@ namespace Aws
             }
 
             g_allocator = nullptr;
-            aws_s3_library_clean_up();
+            #ifdef ENABLE_S3_BUILT
+                aws_s3_library_clean_up();
```
wps132230, line 79: Similarly, `aws_s3_library_clean_up` cleanup several other libraries, including `aws_http_library_clean_up` and `aws_auth_library_clean_up`

source/Api.cpp
sdavtaker: That doesn't seem right :(

PR #289, opened 2021-09-24
Fix iOS Build

ahmedyarub:
*Issue #, if available:*

*Description of changes:*
Keychain methods are not available when building for iOS. They should be required only when building for MacOS and not for all Apple targets.


bretambrose:
I believe this was fixed in https://github.com/awslabs/aws-crt-cpp/pull/322

PR #309, opened 2021-11-15
Hand written doc pass

bretambrose:


include/aws/crt/io/Stream.h
```diff
@@ -43,19 +46,48 @@ namespace Aws
                 InputStream &operator=(InputStream &&) = delete;
 
                 explicit operator bool() const noexcept { return IsValid(); }
+
+                /**
+                 * @return true/false if this object is in a valid state
+                 */
                 virtual bool IsValid() const noexcept = 0;
 
+                /// @private
```
justinboswell, line 55: Shouldn't this be `/** @private */`?
or
```
/**
 * @private
 */

justinboswell (approved):
Fix and ship!

include/aws/crt/io/Stream.h
bretambrose: It looks like it's working to suppress public members as is (see ApiHandle's GetByoCrypto* which don't show up in the current doc set, unlike the Sets)

PR #321, opened 2021-12-12
fix cast compilation error

toge:
*Description of changes:*

On g++ 11.2, explicit type conversion in `TlsOptions.cpp` is caused by compilation error.
So I want to use const_cast.
I hope that this change is in line with the repository policy.


graebm:
Fixed the underlying issue:
https://github.com/awslabs/aws-c-io/pull/456

PR #322, opened 2021-12-22
Fix iOS builds

graebm:
**Issue:**
iOS builds were broken

**Description of Changes:**
- The platform-specific TLS functions have changed so that they're runtime errors now, instead of compile-time errors. This is easier to maintain. See: https://github.com/awslabs/aws-c-io/pull/460
- Run iOS build in CI
  -  Use `build_crt.py` python script to control the cmake args used in CI. The old way of using `builder.json` was accidentally removing cmake args. For example, before this change `-DENABLE_SANITIZERS=ON` was always being removed, and we didn't realize that some tests were actually leaking memory.
- Fix elasticurl, which broke due to changes in the command-line parser, see: https://github.com/awslabs/aws-c-common/pull/878

Also, add iOS build to CI. At this time, the tests aren't run. But at least we can be sure it compiles.

### DIFF
--- a/.builder/actions/build_crt.py
+++ b/.builder/actions/build_crt.py
@@ -1,0 +1,26 @@
+import Builder
+
+
+class BuildCrt(Builder.Action):
+    """
+    Custom build step exists so that the CMake args can be tweaked programmatically.
+    It was proving impossible to tweak them exactly how we wanted via the existing
+    technique of overlaying data from json files.
+    """
+
+    def run(self, env):
+        project = Builder.Project.find_project('aws-crt-cpp')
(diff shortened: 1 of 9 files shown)


.builder/actions/build_crt.py
```diff
@@ -0,0 +1,26 @@
+import Builder
+
+
+class BuildCrt(Builder.Action):
+    """
+    Custom build step exists so that the CMake args can be tweaked programmatically.
```
sbSteveK, line 6: Nice!

sbSteveK (approved):

PR #333, opened 2022-02-14
Connection builder simplification

TwistedTwigleg:
*Description of changes:*

Adds some constructors and singletons to ClientBootstrap, EventLoopGroup, and HostResolver to make the creation of MQTT connections easier and less involved, especially for beginners. This change is loosely modeled after the Java V2 SDK and its code.

This PR contains the following changes:
* Adds singletons to ClientBootstrap, EventLoopGroup, and HostResolver
  * These singletons are modeled to be similar to the Java SDK. As such, each class has two new functions: `GetOrCreateStaticDefault` and `ReleaseStaticDefault`
  * For ClientBootstrap, calling `GetOrCreateStaticDefault` will use the static default singletons for both EventLoopGroup and HostResolver. When calling `ReleaseStaticDefault`, it will call `ReleaseStaticDefault` on both singletons as well.
  * For HostResolver, calling `GetOrCreateStaticDefault` will use the static default EventLoopGroup singleton. When calling `ReleaseStaticDefault`, it will also call `ReleaseStaticDefault` on the EventLoopGroup singleton.
* Adds a new constructor to `WebsocketConfig` and `MqttClient` that do not require passing a ClientBootstrap. If no ClientBootstrap is passed in the constructor, then the static default singleton is used.
* All changes are 100% backwards compatible - existing code is unaffected by anything introduced in this PR (tested by building the samples in the V2 SDK repository both before and after)

_________

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*

### DIFF
--- a/include/aws/crt/Api.h
+++ b/include/aws/crt/Api.h
@@ -125,12 +125,46 @@ namespace Aws
             /// @private
             static const Io::IsTlsAlpnSupportedCallback &GetBYOCryptoIsTlsAlpnSupportedCallback();
 
+            /**
+             * Gets the static default EventLoopGroup, creating it if necessary.
+             *
+             * @return EventLoopGroup* A pointer to the static default EventLoopGroup
+             */
+            static Io::ClientBootstrap *GetOrCreateStaticDefaultClientBootstrap();
+
+            /**
+             * Gets the static default EventLoopGroup, creating it if necessary.
+             *
+             * @return EventLoopGroup* A pointer to the static default EventLoopGroup
+             */
+            static Io::EventLoopGroup *GetOrCreateStaticDefaultEventLoopGroup();
+
+            /**
+             * Gets the static default EventLoopGroup, creating it if necessary.
+             *
+             * @return EventLoopGroup* A pointer to the static default EventLoopGroup
+             */
+            static Io::DefaultHostResolver *GetOrCreateStaticDefaultHostResolver();
+
           private:
             void InitializeLoggingCommon(struct aws_logger_standard_options &options);
 
             aws_logger m_logger;
 
             ApiHandleShutdownBehavior m_shutdownBehavior;
+
+            static Io::ClientBootstrap *s_static_bootstrap;
+            static int s_static_event_loop_group_threads;
+            static Io::EventLoopGroup *s_static_event_loop_group;
+            static int s_host_resolver_default_max_entires;
+            static Io::DefaultHostResolver *s_static_default_host_resolver;
+            static std::mutex s_lock_client_bootstrap;
+            static std::mutex s_lock_event_loop_group;
+            static std::mutex s_lock_default_host_resolver;
+
+            static void ReleaseStaticDefaultClientBootstrap();
+            static void ReleaseStaticDefaultEventLoopGroup();
+            static void ReleaseStaticDefaultHostResolver();
         };
 
         /**
--- a/include/aws/crt/Api.h
+++ b/include/aws/crt/Api.h
@@ -6,9 +6,6 @@
 #include <aws/crt/Types.h>
 #include <aws/crt/crypto/HMAC.h>
 #include <aws/crt/crypto/Hash.h>
-#include <aws/crt/io/Bootstrap.h>
-#include <aws/crt/io/EventLoopGroup.h>
-#include <aws/crt/io/TlsOptions.h>
 #include <aws/crt/mqtt/MqttClient.h>
 
 #include <aws/common/logging.h>
@@ -126,25 +123,40 @@ namespace Aws
             static const Io::IsTlsAlpnSupportedCallback &GetBYOCryptoIsTlsAlpnSupportedCallback();
 
             /**
-             * Gets the static default EventLoopGroup, creating it if necessary.
+             * Gets the static default ClientBootstrap, creating it if necessary.
              *
-             * @return EventLoopGroup* A pointer to the static default EventLoopGroup
+             * @return ClientBootstrap* A pointer to the static default ClientBootstrap
              */
-            static Io::ClientBootstrap *GetOrCreateStaticDefaultClientBootstrap();
+            static Io::ClientBootstrap *GetOrCreateDefaultClientBootstrap();
+
+            /**
+             * Sets the static default ClientBootstrap, if it has not been created already.
+             */
+            static void SetDefaultClientBootstrap(Io::ClientBootstrap *clientBootstrap);
 
             /**
              * Gets the static default EventLoopGroup, creating it if necessary.
              *
              * @return EventLoopGroup* A pointer to the static default EventLoopGroup
              */
-            static Io::EventLoopGroup *GetOrCreateStaticDefaultEventLoopGroup();
+            static Io::EventLoopGroup *GetOrCreateDefaultEventLoopGroup();
 
             /**
-             * Gets the static default EventLoopGroup, creating it if necessary.
-             *
-             * @return EventLoopGroup* A pointer to the static default EventLoopGroup
+             * Sets the static default EventLoopGroup, if it has not been created already.
              */
-            static Io::DefaultHostResolver *GetOrCreateStaticDefaultHostResolver();
+            static void SetDefaultEventLoopGroup(Io::EventLoopGroup *eventLoopGroup);
+
+            /**
+             * Gets the static default HostResolver, creating it if necessary.
+             *
+             * @return EventLoopGroup* A pointer to the static default HostResolver
+             */
+            static Io::HostResolver *GetOrCreateDefaultHostResolver();
+
+            /**
+             * Sets the static default HostResolver, if it has not been created already.
+             */
+            static void SetDefaultHostResolver(Io::HostResolver *hostResolver);
 
           private:
             void InitializeLoggingCommon(struct aws_logger_standard_options &options);
@@ -154,16 +166,16 @@ namespace Aws
             ApiHandleShutdownBehavior m_shutdownBehavior;
 
             static Io::ClientBootstrap *s_static_bootstrap;
-            static int s_static_event_loop_group_threads;
-            static Io::EventLoopGroup *s_static_event_loop_group;
-            static int s_host_resolver_default_max_entires;
-            static Io::DefaultHostResolver *s_static_default_host_resolver;
             static std::mutex s_lock_client_bootstrap;
-            static std::mutex s_lock_event_loop_group;
-            static std::mutex s_lock_default_host_resolver;
-
             static void ReleaseStaticDefaultClientBootstrap();
+
+            static Io::EventLoopGroup *s_static_event_loop_group;
+            static std::mutex s_lock_event_loop_group;
             static void ReleaseStaticDefaultEventLoopGroup();
+
+            static int s_host_resolver_default_max_entries;
+            static Io::HostResolver *s_static_default_host_resolver;
+            static std::mutex s_lock_default_host_resolver;
             static void ReleaseStaticDefaultHostResolver();
         };
 
--- a/include/aws/crt/Api.h
+++ b/include/aws/crt/Api.h
@@ -135,12 +135,7 @@ namespace Aws
              *
              * @return ClientBootstrap* A pointer to the static default ClientBootstrap
              */
-            static Io::ClientBootstrap *GetOrCreateDefaultClientBootstrap();
-
-            /**
-             * Sets the static default ClientBootstrap, if it has not been created already.
-             */
-            static void SetDefaultClientBootstrap(Io::ClientBootstrap *clientBootstrap);
+            static Io::ClientBootstrap *GetOrCreateStaticDefaultClientBootstrap();
 
             /**
              * Gets the static default EventLoopGroup, creating it if necessary.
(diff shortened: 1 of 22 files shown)


bretambrose (changes requested):
Quick first glance feedback:

* Use Aws::New/Delete over new/delete
* All of the static manipulation needs to be threadsafe.
* ApiHandle cleanup should release the statics

include/aws/crt/Api.h
```diff
@@ -125,12 +125,46 @@ namespace Aws
...
             void InitializeLoggingCommon(struct aws_logger_standard_options &options);
 
             aws_logger m_logger;
 
             ApiHandleShutdownBehavior m_shutdownBehavior;
+
+            static Io::ClientBootstrap *s_static_bootstrap;
+            static int s_static_event_loop_group_threads;
+            static Io::EventLoopGroup *s_static_event_loop_group;
+            static int s_host_resolver_default_max_entires;
```
bretambrose, line 159: entries

include/aws/crt/Api.h
```diff
@@ -125,12 +125,46 @@ namespace Aws
...
+
           private:
             void InitializeLoggingCommon(struct aws_logger_standard_options &options);
 
             aws_logger m_logger;
 
             ApiHandleShutdownBehavior m_shutdownBehavior;
+
+            static Io::ClientBootstrap *s_static_bootstrap;
+            static int s_static_event_loop_group_threads;
```
bretambrose, line 157: Minor but I'd group these by subject (bootstrap, elg, host resolver)

include/aws/crt/Api.h
```diff
@@ -125,12 +125,46 @@ namespace Aws
             /// @private
             static const Io::IsTlsAlpnSupportedCallback &GetBYOCryptoIsTlsAlpnSupportedCallback();
 
+            /**
```
bretambrose, line 128: It's not your changes, but the includes on this header file are way too much.  Maybe see what can be removed (with appropriate forward declarations).

source/io/Bootstrap.cpp
```diff
@@ -2,6 +2,7 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
+#include <aws/crt/Types.h>
```
bretambrose, line 5: Why was this needed?

source/io/EventLoopGroup.cpp
```diff
@@ -2,8 +2,11 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
+#include <aws/crt/Types.h>
```
bretambrose, line 5: Feels like all the new includes should be removable given the migration to ApiHandle

source/io/HostResolver.cpp
```diff
@@ -2,11 +2,11 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
-#include <aws/crt/io/HostResolver.h>
-
-#include <aws/crt/io/EventLoopGroup.h>
-
 #include <aws/common/string.h>
```
bretambrose, line 5: Seems like this should be revertable.

bretambrose (approved):
Fix n ship

include/aws/crt/Api.h
```diff
@@ -125,12 +125,46 @@ namespace Aws
...
+             * @return EventLoopGroup* A pointer to the static default EventLoopGroup
+             */
+            static Io::EventLoopGroup *GetOrCreateStaticDefaultEventLoopGroup();
+
+            /**
+             * Gets the static default EventLoopGroup, creating it if necessary.
+             *
+             * @return EventLoopGroup* A pointer to the static default EventLoopGroup
+             */
+            static Io::DefaultHostResolver *GetOrCreateStaticDefaultHostResolver();
```
graebm, line 147: trivial: comments say EventLoopGroup, but this is HostResolver

include/aws/crt/Api.h
```diff
@@ -125,12 +125,46 @@ namespace Aws
             /// @private
             static const Io::IsTlsAlpnSupportedCallback &GetBYOCryptoIsTlsAlpnSupportedCallback();
 
+            /**
+             * Gets the static default EventLoopGroup, creating it if necessary.
+             *
+             * @return EventLoopGroup* A pointer to the static default EventLoopGroup
+             */
+            static Io::ClientBootstrap *GetOrCreateStaticDefaultClientBootstrap();
```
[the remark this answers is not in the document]
graebm: trivial: also, copy/paste error in comments. says "EventLoopGroup" but this is "ClientBootstrap"

include/aws/crt/Api.h
graebm, line 147: Return the base class type `HostResolver` not the specific subclass `DefaultHostResolver`

include/aws/iot/MqttClient.h
```diff
@@ -120,6 +120,15 @@ namespace Aws
                 Crt::Io::ClientBootstrap *bootstrap,
                 Crt::Allocator *allocator = Crt::g_allocator) noexcept;
 
+            /**
+             * Create a websocket configuration for use with the default credentials provider chain and default
+             * ClientBootstrap. Signing region will be used for Sigv4 signature calculations.
```
graebm, line 125: Try and get doxygen to do a link to the original `ApiHandle::GetOrCreateDefaultXYZ()` function. Where we should have more documentation about "default XYZ"

you can update the docs by installing Doxygen and running
```
cd docsrc
doxygen Doxyfile
```

then you can look at the webpages that appear under the `docs/` folder locally to see if they look good

include/aws/crt/Api.h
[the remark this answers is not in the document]
graebm: trivial: Add comments about how this default will be used in places where the user doesn't explicitly pass a pointer blah blah. Comment about it's lifetime. Make it a nice informative summary of everything the user might want to know about "default XYZ", then have all places that use the "default XYZ" link back to this function, where the authorative.

Repeat these similar comments for each `GetOrCreateDefaultXYZ()` function. It's fine to be a bit repetitive in docs, because users don't read them top to bottom.

source/auth/Credentials.cpp
```diff
@@ -265,7 +267,9 @@ namespace Aws
                 struct aws_credentials_provider_chain_default_options raw_config;
                 AWS_ZERO_STRUCT(raw_config);
 
-                raw_config.bootstrap = config.Bootstrap->GetUnderlyingHandle();
+                raw_config.bootstrap =
+                    config.Bootstrap ? config.Bootstrap->GetUnderlyingHandle()
+                                     : ApiHandle::GetOrCreateStaticDefaultClientBootstrap()->GetUnderlyingHandle();
```
graebm, line 272: You need to add documentation to `CredentialsProviderChainDefaultConfig::Bootstrap` about how the default will be used if it's not set.

source/auth/Credentials.cpp
```diff
@@ -279,7 +283,9 @@ namespace Aws
                 struct aws_credentials_provider_x509_options raw_config;
                 AWS_ZERO_STRUCT(raw_config);
 
-                raw_config.bootstrap = config.Bootstrap->GetUnderlyingHandle();
+                raw_config.bootstrap =
+                    config.Bootstrap ? config.Bootstrap->GetUnderlyingHandle()
```
graebm, line 287: add docs to `CredentialsProviderX509Config::Bootstrap` about defaults

source/auth/Credentials.cpp
```diff
@@ -279,7 +283,9 @@ namespace Aws
...
+                                     : ApiHandle::GetOrCreateStaticDefaultClientBootstrap()->GetUnderlyingHandle();
                 raw_config.tls_connection_options = config.TlsOptions.GetUnderlyingHandle();
                 raw_config.thing_name = aws_byte_cursor_from_c_str(config.ThingName.c_str());
                 raw_config.role_alias = aws_byte_cursor_from_c_str(config.RoleAlias.c_str());
```
graebm, line 291: I would like us to completely scrub aws-crt-cpp for all places where EventLoopGroup|HostResolver|ClientBootstrap get passed around, and make it so they ALL work with the new GetOrCreateDefaultXYZ() functions.

So we'd be looking for constructors/creation-functions that take these as args, and for "config/options" classes that have these as members.

Here's me using regex in vscode to search all header files where we might be passing one of these around:
Screen Shot 2022-02-24 at 3 43 03 PM

include/aws/crt/Api.h
```diff
@@ -125,12 +125,46 @@ namespace Aws
...
+
+            /**
```
graebm, line 135: Debatable: personally, I'd also like to have `void SetDefaultEventLoopGroup(Io::EventLoopGroup *eventLoopGroup)` so the user could create one with the exact number of threads they desire, set it at program startup, and not worry about passing it around explicitly.

This is debatable. We can add these GetOrCreateDefault functions now and SetDefault functions later. But like, we're doing this now so it's kind of the time...

source/iot/MqttClient.cpp
```diff
@@ -45,6 +45,33 @@ namespace Aws
...
         }
 
+        WebsocketConfig::WebsocketConfig(const Crt::String &signingRegion, Crt::Allocator *allocator) noexcept
+            : SigningRegion(signingRegion), ServiceName("iotdevicegateway")
+        {
+            Crt::Auth::CredentialsProviderChainDefaultConfig config;
+            config.Bootstrap = Crt::ApiHandle::GetOrCreateStaticDefaultClientBootstrap();
+
+            CredentialsProvider =
+                Crt::Auth::CredentialsProvider::CreateCredentialsProviderChainDefault(config, allocator);
```
graebm, line 55: No need to copy/paste so much code. You can have this new constructor call into another one like:
```
WebsocketConfig::WebsocketConfig(const Crt::String &signingRegion, Crt::Allocator *allocator) noexcept
    : WebsocketConfig(signingRegion, ApiHandle::GetOrCreateDefaultClientBootstrap(), allocator)
{}
```

source/iot/MqttClient.cpp
```diff
@@ -389,6 +416,15 @@ namespace Aws
             }
         }
 
+        MqttClient::MqttClient(Crt::Allocator *allocator) noexcept
+            : m_client(*Crt::ApiHandle::GetOrCreateStaticDefaultClientBootstrap(), allocator), m_lastError(0)
+        {
+            if (!m_client)
+            {
+                m_lastError = m_client.LastError();
```
graebm, line 424: same, just have this constructor call into the pre-existing one

include/aws/crt/Api.h
```diff
@@ -125,12 +125,46 @@ namespace Aws
...
             ApiHandleShutdownBehavior m_shutdownBehavior;
+
+            static Io::ClientBootstrap *s_static_bootstrap;
+            static int s_static_event_loop_group_threads;
+            static Io::EventLoopGroup *s_static_event_loop_group;
+            static int s_host_resolver_default_max_entires;
+            static Io::DefaultHostResolver *s_static_default_host_resolver;
+            static std::mutex s_lock_client_bootstrap;
+            static std::mutex s_lock_event_loop_group;
+            static std::mutex s_lock_default_host_resolver;
```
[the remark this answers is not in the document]
TwistedTwigleg: Thanks, but I cannot take credit for it - I forgot it initially and added it after a code review from Bret 🙂

source/auth/Credentials.cpp
TwistedTwigleg: Should this be a separate PR/ticket? I don't mind making the change, but I'm wondering since, on the surface, it seems like it might be a bigger change.

include/aws/crt/auth/Credentials.h
```diff
@@ -284,6 +293,9 @@ namespace Aws
                 /**
                  * Connection bootstrap to use to create the http connection required to
                  * query credentials from the x509 provider
+                 *
+                 * Note: If a ClientBootstrap is not passed in Bootstrap, then the default ClientBootstrap in Api will
+                 * be used automatically.
```
graebm, line 298: trivial: same, include a link to docs about default

source/auth/Credentials.cpp
graebm: Still need
- ImdsClientConfig::Bootstrap
- CredentialsProviderImdsConfig::Bootstrap
- HttpClientConnectionOptions::Bootstrap

source/Api.cpp
```diff
@@ -59,6 +69,10 @@ namespace Aws
 
         ApiHandle::~ApiHandle()
         {
+            ReleaseStaticDefaultClientBootstrap();
+            ReleaseStaticDefaultEventLoopGroup();
+            ReleaseStaticDefaultHostResolver();
```
graebm, line 74: Add a test to ensure that, when these default objects get automatically created, their memory gets cleaned up. Could probably be as simple as a test that just creates the ClientBootstrap (resulting in the other 2 getting created)

bretambrose (approved):

TwistedTwigleg:
Thanks for the reviews! Merging into `main`...

PR #341, opened 2022-03-02
adapt new input stream api

TingDaoK:
*Issue #, if available:*

*Description of changes:*


include/aws/crt/RefCounted.h
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Aws
+{
+    namespace Crt
+    {
+        /**
+         * Inherit from RefCounted to allow reference-counting from C code,
+         * which will keep your C++ object alive as long as the count is non-zero.
+         *
+         * A class must inherit from RefCounted and enable_shared_from_this.
```
graebm, line 18: trivial: std::enable_shared_from_this

source/io/Stream.cpp
```diff
@@ -68,18 +68,33 @@ namespace Aws
...
 
-            void InputStream::s_Destroy(struct aws_input_stream *stream)
+            aws_input_stream *InputStream::s_Acquire(aws_input_stream *stream)
             {
-                (void)stream;
-                // DO NOTHING, let the C++ destructor handle it.
+                if (stream)
+                {
+                    auto impl = static_cast<InputStream *>(stream->impl);
+                    impl->AcquireRef();
```
graebm, line 76: this can be simplified to be just these 2 lines

source/io/Stream.cpp
```diff
@@ -68,18 +68,33 @@ namespace Aws
...
+                }
+                return stream;
+            }
+
+            aws_input_stream *InputStream::s_Release(aws_input_stream *stream)
+            {
+                if (stream)
+                {
+                    auto impl = static_cast<InputStream *>(stream->impl);
+                    impl->ReleaseRef();
```
graebm, line 86: same

graebm (approved):

graebm (commented):
remove anywhere we have c++ keeping the InputStream alive. we don't need that complexity now that we're refcounting from C.

Also, add a test to confirm that the RefCounted class works like we think it does.
Confirm that C can keep a C++ InputStream instance alive. Confirm that C refcount can go to 0, then back to 1, and things still work

PR #376, opened 2022-05-10
fix: make static build compiler flags public so they are inherited

MikeDombo:
*Issue #, if available:*

*Description of changes:*
Without this change, compiling with `-DSTATIC_CRT=ON` does not statically compile the subprojects (Greengrass IPC for example).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -284,9 +284,9 @@ aws_prepare_symbol_visibility_args(${PROJECT_NAME} "AWS_CRT_CPP")
 #set runtime library
 if (MSVC)
     if(STATIC_CRT)
-        target_compile_options(${PROJECT_NAME} PRIVATE "/MT$<$<CONFIG:Debug>:d>")
+        target_compile_options(${PROJECT_NAME} PUBLIC "/MT$<$<CONFIG:Debug>:d>")
     else()
-        target_compile_options(${PROJECT_NAME} PRIVATE "/MD$<$<CONFIG:Debug>:d>")
+        target_compile_options(${PROJECT_NAME} PUBLIC "/MD$<$<CONFIG:Debug>:d>")
     endif()
 endif ()
 


bretambrose (approved):

CMakeLists.txt
```diff
@@ -284,9 +284,9 @@ aws_prepare_symbol_visibility_args(${PROJECT_NAME} "AWS_CRT_CPP")
 #set runtime library
 if (MSVC)
     if(STATIC_CRT)
-        target_compile_options(${PROJECT_NAME} PRIVATE "/MT$<$<CONFIG:Debug>:d>")
+        target_compile_options(${PROJECT_NAME} PUBLIC "/MT$<$<CONFIG:Debug>:d>")
```
graebm, line 287: Ugh. I feel both ways about this change. Mildly against. I agree this makes it simpler for people to get a project building. But it seems like we're trying to make our library fix an issue that's not ours to fix. It sucks MSVC forces every library in a project needs to be built exactly the same way. But my understanding is that it's not good practice for compiler options to be PUBLIC in CMake (Please find me a good example, I've searched and all I find is 100 pages explaining what PUBLIC means, not when to use it). We've had bugs in the past where we accidentally made things PUBLIC and it was messing with people consuming our library.

Here's research I did looking deeper into this...

CMake 3.15 introduced [CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html). But this is just something the user passes at configure time like `-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded`. But it only affects the one project. It's still up to the user to make sure every other library they build, and the final application, are also configured this way. We could support this new `CMAKE_MSVC_RUNTIME_LIBRARY` variable, but it would either be a breaking change, or a ton of code to try and support both `STATIC_CRT` and `CMAKE_MSVC_RUNTIME_LIBRARY`.

graebm, line 287: Ugh. I feel both ways about this change. Mildly against. I agree this makes it simpler for people to get a project building. But it seems like we're trying to make our library fix an issue that's not ours to fix. It sucks MSVC forces every library in a project needs to be built exactly the same way. But my understanding is that it's not good practice for compiler options to be PUBLIC in CMake (Please find me a good example, I've searched and all I find is 100 pages explaining what PUBLIC means, not when to use it). We've had bugs in the past where we accidentally made things PUBLIC and it was messing with people consuming our library.

Here's research I did looking deeper into this...

CMake 3.15 introduced [CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html). But this is just something the user passes at configure time like `-DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded`. But it only affects the one project. It's still up to the user to make sure every other library they build, and the final application, are also configured this way. We could support this new `CMAKE_MSVC_RUNTIME_LIBRARY` variable, but it would either be a breaking change, or a ton of code to try and support both `STATIC_CRT` and `CMAKE_MSVC_RUNTIME_LIBRARY`.

Issue #388, opened 2022-07-22
[ClientBootstrap]: segfault at program exit


grrtrr:
### Summary

This issue happens regardless of whether `ClientBootstrap::EnableBlockingShutdown()` has been called, and it is caused by returning from `~ClientBootstrap()` while asynchronous shutdown is still in progress.

In production, we are repeatedly getting the same backtrace:
```console
FATAL: Received signal 11 (Segmentation fault)
Backtrace (most recent call first)
#6	 <?> at 0x564a57ceca28 in Aws::Utils::Logging::s_aws_logger_redirect_log()
#5	 <?> at 0x564a57d4bb2b in aws_default_dns_resolve
#4	 <?> at 0x564a57d4dfed in resolver_thread_fn
#3	 <?> at 0x564a57d5a117 in thread_fn
#2	 <?> at 0x564a5753f32c in thread_metrics_start_routine
#1	 <?> at 0x7f04dcfb86db in start_thread
```

### Details

The orderly shutdown of the client bootstrap is invoked by calling `CleanupCrt()`, followed by shutting down CRT logging (see 
https://github.com/aws/aws-sdk-cpp/issues/1995).
```c++
// aws core/source/Globals.cpp
    void CleanupCrt()
    {
        Aws::SetDefaultClientBootstrap(nullptr);
        // ...
    }
```

The `s_client_bootstrap_destroy_impl` invokes the "blocking shutdown" callback right after the `aws_event_loop_group_release` and `aws_host_resolver_release` calls:
```c
static void s_client_bootstrap_destroy_impl(struct aws_client_bootstrap *bootstrap) {
    AWS_ASSERT(bootstrap);
    AWS_LOGF_DEBUG(AWS_LS_IO_CHANNEL_BOOTSTRAP, "id=%p: destroying", (void *)bootstrap);
    aws_client_bootstrap_shutdown_complete_fn *on_shutdown_complete = bootstrap->on_shutdown_complete;
    void *user_data = bootstrap->user_data;

    aws_event_loop_group_release(bootstrap->event_loop_group);
    aws_host_resolver_release(bootstrap->host_resolver);

    if (on_shutdown_complete) {
        on_shutdown_complete(user_data);
    }
}
```

However, both the `event_loop_group` and the `host_resolver` perform _asynchronous shutdown_, i.e. their "destroy" functions merely trigger the shutdown of threads that are still running; they do not await thread exit.

[...]

PR #389, opened 2022-07-22
[ClientBootstrap] await thread shutdown (#388)

grrtrr:
*Issue #, if available:* #388

*Description of changes:*
The SDK calls `CleanupCrt()` followed by `ShutdownCRTLogging()`. The former starts destruction of the client bootstrap.
The client bootstrap currently performs _asynchronous shutdown_, which means that exiting threads may attempt to write to the CRT logging subsystem after it has been deleted, causing segmentation faults.

Avoid asynchronous shutdown by awaiting the exiting of the `event_loop_group` and `host_resolver` threads.

### DIFF
--- a/source/io/Bootstrap.cpp
+++ b/source/io/Bootstrap.cpp
@@ -2,8 +2,13 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
-#include <aws/crt/Api.h>
+#include <aws/common/clock.h>
+
 #include <aws/crt/io/Bootstrap.h>
+
+// Maximum time to wait (in seconds) for the blocking shutdown sequence.
+// This is to avoid hanging at program exit due to unexpected conditions.
+#define FINAL_TIMEOUT_SECS 10
 
 namespace Aws
 {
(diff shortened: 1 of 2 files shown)


source/io/Bootstrap.cpp
```diff
@@ -2,75 +2,88 @@
...
-
-                    callbackData->ShutdownPromise.set_value();
-                    if (callbackData->ShutdownCallback)
-                    {
-                        callbackData->ShutdownCallback();
-                    }
-
-                    Crt::Delete(callbackData, callbackData->m_allocator);
+                    const auto shutdown_user_data = static_cast<struct shutdown_user_data *>(pred_ctx);
+                    return shutdown_user_data->shutdown_count == 3;
```
ujtardiff, line 38: Hmmm, is there an alternative strategy to counting to 3 in the shutdown callbacks ? I'm wondering if something more explicit can help, because if a developer adds another system to shut down, are they going to know to setup the shutdown callback and set this check to 4 ?

grrtrr: Passing in the same callback/data into all three `aws_shutdown_callback_options` simplifies the implementation.
Setting each individual completion in a callback would require 3 different types of callbacks, each setting a different condition.
This would not solve the problem if someone modifies `aws_client_bootstrap` to include more asynchronous shutdown operations; they would still have to update the code.

grrtrr:
This was a red herring. See comments in #389.

PR #393, opened 2022-08-12
Fix to cmake detection of CRT version by consumers

sdavtaker:
- Adding support for config-version.cmake generation at the top level CMakeLists.txt
- Stripping starting v on version numbers on the update_version script

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,9 +347,24 @@ configure_file("cmake/${PROJECT_NAME}-config.cmake"
         "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}-config.cmake"
         @ONLY)
 
+include(CMakePackageConfigHelpers)
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 if (NOT DEFINED AWS_CRT_CPP_VERSION)
-    set(AWS_CRT_CPP_VERSION "v0.17.30")
+    set(AWS_CRT_CPP_VERSION "0.17.30")
```
graebm, line 73: Don't strip the "v" off here, this string is already being used for metrics
https://github.com/awslabs/aws-crt-cpp/blob/421a69a81fb73b4c9917ffb3d095c2c65e9e973d/source/iot/MqttClient.cpp#L350

Make a new variable that strips off the "v" and use that for package version

sdavtaker: Fixed, published new version.

graebm (approved):

PR #409, opened 2022-09-19
Cognito provider

bretambrose:
* Cognito credentials provider binding
* STS credentials provider binding

### DIFF
--- a/include/aws/crt/auth/Credentials.h
+++ b/include/aws/crt/auth/Credentials.h
@@ -336,6 +336,71 @@ namespace Aws
             };
 
             /**
+             * A pair defining an identity provider and a valid login token sourced from it.
+             */
+            struct AWS_CRT_CPP_API CognitoLoginPair
+            {
+
+                /**
+                 * Name of an identity provider
+                 */
+                String IdentityProviderName;
+
+                /**
+                 * Valid login token source from the identity provider
+                 */
+                String IdentityProviderToken;
+            };
+
+            /**
+             * Configuration options for the Cognito credentials provider
+             */
+            struct AWS_CRT_CPP_API CredentialsProviderCognitoConfig
+            {
+                CredentialsProviderCognitoConfig();
+
+                /**
+                 * Cognito service regional endpoint to source credentials from.
+                 */
+                String Endpoint;
+
+                /**
+                 * Cognito identity to fetch credentials relative to.
+                 */
+                String Identity;
+
+                /**
+                 * Optional set of identity provider token pairs to allow for authenticated identity access.
+                 */
+                Optional<Vector<CognitoLoginPair>> Logins;
+
+                /**
+                 * Optional ARN of the role to be assumed when multiple roles were received in the token from the
+                 * identity provider.
+                 */
+                Optional<String> CustomRoleArn;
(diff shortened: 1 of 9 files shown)


include/aws/crt/auth/Credentials.h
```diff
@@ -335,6 +335,118 @@ namespace Aws
...
+                /**
+                 * How long sourced credentials should remain valid for, in seconds.  900 is the minimum allowed value.
+                 */
+                uint16_t DurationSeconds;
+
+                /**
+                 * Connection bootstrap to use to create the http connection required to
+                 * query credentials from the STS provider
+                 *
+                 * Note: If null, then the default ClientBootstrap is used
```
TwistedTwigleg, line 434: Super minor nitpick: `null` -> `nullptr`

include/aws/crt/auth/Credentials.h
```diff
@@ -448,6 +560,20 @@ namespace Aws
                     const CredentialsProviderDelegateConfig &config,
                     Allocator *allocator = ApiAllocator());
 
+                /**
+                 * Creates a provider that sources credentials from the Cognito Identity service
```
TwistedTwigleg, line 564: Super minor nitpick: Should `Cognito Identity service` be `Cognito credentials provider`? Or are they separate things?

source/auth/Credentials.cpp
```diff
@@ -372,6 +372,98 @@ namespace Aws
...
+            CredentialsProviderSTSConfig::CredentialsProviderSTSConfig() : Bootstrap(nullptr) {}
+
+            std::shared_ptr<ICredentialsProvider> CredentialsProvider::CreateCredentialsProviderSTS(
+                const CredentialsProviderSTSConfig &config,
+                Allocator *allocator)
+            {
+                struct aws_credentials_provider_sts_options raw_config;
+                AWS_ZERO_STRUCT(raw_config);
+
+                raw_config.creds_provider = config.Provider->GetUnderlyingHandle();
```
TwistedTwigleg, line 444: Question: Do we need to check if a provider is defined?

TwistedTwigleg (approved):
Looks good! There was a few minor nitpicks and a question, but they are minor and it looks good to me 👍

include/aws/crt/auth/Credentials.h
bretambrose: While nullptr is the C++11 syntax, colloquially I always hear "null" (is-null, is-not-null, etc...)  It feels really weird and awkward to me the other way.

include/aws/crt/auth/Credentials.h
bretambrose: The Cognito-Identity is the web service that the provider sources from, so it should be good as written.

source/auth/Credentials.cpp
bretambrose: Yeah, that seems like a good check.

include/aws/crt/auth/Credentials.h
TwistedTwigleg: That is fair. In that case, then I'd leave it as `null` 👍

PR #419, opened 2022-10-27
Endpoint resolution engine bindings

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

Endpoint resolution engine bindings


include/aws/crt/endpoints/RuleEngine.h
```diff
@@ -0,0 +1,157 @@
...
+             *   endpoint could not be resolved.
+             */
+            class AWS_CRT_CPP_API ResolutionOutcome final
+            {
+              public:
+                ~ResolutionOutcome();
+
+                /* TODO: move/copy semantics */
+                ResolutionOutcome(const ResolutionOutcome &) = delete;
+                ResolutionOutcome &operator=(const ResolutionOutcome &) = delete;
```
SergeyRyabinin, line 86: very minor nit
why & next to operator not the type?

source/endpoints/RuleEngine.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+            Optional<StringView> ResolutionOutcome::GetUrl() const noexcept
+            {
+                ByteCursor url;
+                if (aws_endpoints_resolved_endpoint_get_url(m_resolvedEndpoint, &url))
+                {
+                    return Optional<StringView>();
+                }
+
+                return Optional<StringView>(ByteCursorToStringView(url));
```
SergeyRyabinin, line 75: both c-tors with arguments of Crt::Optional
```
Optional(const T &val)
Optional(T &&val)
```
have a call to `new`
and `new` can throw at least std::bad_alloc

source/endpoints/RuleEngine.cpp
```diff
@@ -0,0 +1,177 @@
...
+            }
+
+            inline StringView CrtStringToStringView(const aws_string *s)
+            {
+                ByteCursor key = aws_byte_cursor_from_string(s);
+                return ByteCursorToStringView(key);
+            }
+
+            Optional<UnorderedMap<StringView, Vector<StringView>>> ResolutionOutcome::GetHeaders() const noexcept
```
SergeyRyabinin, line 84: this function is declared as `noexcept`
while it obviously allocate memory using `new` in both Optional and UnorderedMap and Vector.

include/aws/crt/endpoints/RuleEngine.h
DmitriyMusatkin: dont know the history of this, but this is the format used across the package, so just staying consistent

source/endpoints/RuleEngine.cpp
```diff
@@ -0,0 +1,177 @@
...
+                    return Optional<UnorderedMap<StringView, Vector<StringView>>>();
+                }
+
+                UnorderedMap<StringView, Vector<StringView>> headers;
+                for (struct aws_hash_iter iter = aws_hash_iter_begin(resolved_headers); !aws_hash_iter_done(&iter);
+                     aws_hash_iter_next(&iter))
+                {
+                    ByteCursor key = aws_byte_cursor_from_string((const aws_string *)iter.element.key);
+                    const aws_array_list *array = (const aws_array_list *)iter.element.value;
+                    headers.insert(
```
SergeyRyabinin: although `insert` may take the most efficient copy-elision overload,
have you tried/considered `emplace` to be sure that copy of a temporary object does not happen?

source/endpoints/RuleEngine.cpp
DmitriyMusatkin: hmm, good point. i see we are doing noexcept in other cases where we create optional (ex. http classes). will remove noexcept

source/endpoints/RuleEngine.cpp
DmitriyMusatkin: yeah, emplace would be better here.

SergeyRyabinin (approved):

TingDaoK (approved):
LGTM

PR #441, opened 2022-12-02
Fix destruct http message double free

waahm7:
*Description of changes:*
- This fixes an issue that aws_input_stream was being freed twice.

### DIFF
--- a/source/http/HttpRequestResponse.cpp
+++ b/source/http/HttpRequestResponse.cpp
@@ -26,21 +26,7 @@ namespace Aws
                 }
             }
 
-            HttpMessage::~HttpMessage()
-            {
-                if (m_message != nullptr)
-                {
-                    aws_input_stream *old_stream = aws_http_message_get_body_stream(m_message);
-                    if (old_stream != nullptr)
-                    {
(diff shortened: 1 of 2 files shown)


source/http/HttpRequestResponse.cpp
```diff
@@ -26,21 +26,7 @@ namespace Aws
...
-                    {
-                        aws_input_stream_destroy(old_stream);
-                    }
-
-                    aws_http_message_release(m_message);
-
-                    m_message = nullptr;
-                }
-            }
+            HttpMessage::~HttpMessage() { m_message = aws_http_message_release(m_message); }
```
graebm, line 29: let's add a regression test to repro the issue, and prove it's fixed

waahm7: Thanks, I have added an assert which makes sure that ref count can not be -ve. So the following test started failing because of double free.
- HttpRequestTestCreateDestroy
- Sigv4SigningTestSimple 
- Sigv4SigningTestCredentials
- Sigv4SigningTestUnsignedPayload

They are working now.

graebm (approved):

PR #455, opened 2023-01-19
Mqtt311 operation statistics support

TwistedTwigleg:
*Description of changes:*

Adds operation statistics support to `aws-crt-nodejs` for MQTT311, as well as fixing a segfault with operation statistics support in MQTT5. Also adds operation statistics support tests for both MQTT311 and MQTT5 to increase code coverage and safety.

~~Relies on https://github.com/awslabs/aws-c-mqtt/pull/247 being merged and released.~~

__________


source/mqtt/MqttClient.cpp
```diff
@@ -698,6 +700,21 @@ namespace Aws
                 return packetId;
             }
 
+            const MqttConnectionOperationStatistics &MqttConnection::GetOperationStatistics() noexcept
+            {
+                if (m_underlyingConnection != nullptr)
+                {
+                    aws_mqtt_client_connection_get_stats(m_underlyingConnection, &m_operationStatisticsNative);
```
xiazhvera, line 707: A minor suggestions is that I feel it is not necessary to make the operationStatisticsNative as a class member as we only use the data once here. 
We could do something like:
```
aws_mqtt5_client_operation_statistics operationStatisticsNative = {0,0,0,0};
aws_mqtt_client_connection_get_stats(m_underlyingConnection, &operationStatisticsNative);
```

xiazhvera (approved):
Added a suggestion in the review, other than that it looks good to me. 
Thanks for fixing the bug in Mqtt5 GetOperationStatistics!

source/mqtt/MqttClient.cpp
TwistedTwigleg: Good idea! I'll adjust it

TwistedTwigleg:

PR #458, opened 2023-01-24
Fix Compiler Warnings

waahm7:
*Description of changes:*
- Always treat warnings as error irrespective of debug string
- Fixes unused allocator variable in two places.
- No need to handle error `InitDefaultClient` as that can't fail.
- Handle error properly for TlsContextOptions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,7 +338,7 @@ endif()
 if(MSVC)
     target_compile_options(${PROJECT_NAME} PRIVATE /W4 /WX /wd4068)
 else()
-    target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wextra -Wno-long-long -pedantic -Werror)
+    target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wno-long-long -pedantic -Werror)
 endif()
 
 target_include_directories(${PROJECT_NAME} PUBLIC
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,10 +335,12 @@ if(CMAKE_BUILD_TYPE STREQUAL "" OR CMAKE_BUILD_TYPE MATCHES Debug)
 endif()
 
  # set extra warning flags
-if(MSVC)
-    target_compile_options(${PROJECT_NAME} PRIVATE /W4 /WX /wd4068)
-else()
-    target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wno-long-long -pedantic -Werror)
+if(AWS_WARNINGS_ARE_ERRORS)
+    if(MSVC)
+        target_compile_options(${PROJECT_NAME} PRIVATE /W4 /WX /wd4068)
+    else()
+        target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wno-long-long -pedantic -Werror)
+endif()
 endif()
 
 target_include_directories(${PROJECT_NAME} PUBLIC
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,13 +332,13 @@ endif()
 
 if(CMAKE_BUILD_TYPE STREQUAL "" OR CMAKE_BUILD_TYPE MATCHES Debug)
     target_compile_definitions(${PROJECT_NAME} PRIVATE "-DDEBUG_BUILD")
+endif()
 
-    # set extra warning flags for debug build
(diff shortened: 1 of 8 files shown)


SergeyRyabinin (approved):

source/external/cJSON.cpp
```diff
@@ -426,7 +426,7 @@ CJSON_PUBLIC(char*) cJSON_SetValuestring(cJSON *object, const char *valuestring)
     }
     if (strlen(valuestring) <= strlen(object->valuestring))
     {
-        strncpy(object->valuestring, valuestring, strlen(valuestring) + 1);
```
waahm7, line 429: This causes 
```
error: char* strncpy(char*, const char*, size_t) specified bound depends on the length of the source argument [-Werror=stringop-truncation]
  429 |         strncpy(object->valuestring, valuestring, strlen(valuestring) + 1);
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/root/aws-crt-cpp/source/external/cJSON.cpp:427:15: note: length computed here
  427 |     if (strlen(valuestring) <= strlen(object->valuestring))
  ```

File: include/aws/crt/mqtt/Mqtt5Packets.h [hunk 2]
```diff
@@ -448,10 +448,7 @@ namespace Aws
             class AWS_CRT_CPP_API NegotiatedSettings
             {
               public:
-                NegotiatedSettings(
-                    const aws_mqtt5_negotiated_settings &negotiated_settings,
-
-                    Allocator *allocator = ApiAllocator()) noexcept;
+                NegotiatedSettings(const aws_mqtt5_negotiated_settings &negotiated_settings) noexcept;
```
graebm, line 452: Don't change this. The pattern is that everything takes a custom allocator. This datastructure just happens to not need it at the moment

instead, go into the .cpp file and mark it unused via `(void)allocator;`

include/aws/crt/mqtt/Mqtt5Packets.h
```diff
@@ -2209,9 +2206,7 @@ namespace Aws
             class AWS_CRT_CPP_API UnSubAckPacket : public IPacket
             {
               public:
-                UnSubAckPacket(
-                    const aws_mqtt5_packet_unsuback_view &packet,
-                    Allocator *allocator = ApiAllocator()) noexcept;
+                UnSubAckPacket(const aws_mqtt5_packet_unsuback_view &packet) noexcept;
```
graebm, line 2209: same

source/external/cJSON.cpp
TwistedTwigleg: Question: Should this be `size_t value_length = strlen(valuestring) + sizeof("");` like in aws-c-common: https://github.com/awslabs/aws-c-common/blob/3973e3c89bbb532f5c3d2327b602836a57bd4db6/source/external/cJSON.c#L410?

CMakeLists.txt
```diff
@@ -332,13 +332,13 @@ endif()
...
-    if(MSVC)
-        target_compile_options(${PROJECT_NAME} PRIVATE /W4 /WX /wd4068)
-    else()
-        target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wno-long-long -pedantic -Werror)
-    endif()
+ # set extra warning flags
+if(MSVC)
+    target_compile_options(${PROJECT_NAME} PRIVATE /W4 /WX /wd4068)
+else()
+    target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wno-long-long -pedantic -Werror)
```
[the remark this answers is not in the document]
waahm7: Thanks, makes sense.

source/external/cJSON.cpp
waahm7: I think `sizeof("")` is always 1. So it doesn't make a difference. https://stackoverflow.com/questions/4141666/why-sizeof-is-equivalent-to-1-and-sizeofnull-is-equivalent-to-4-in-c-langu

TwistedTwigleg: > Not related, but should we remove the copy of cJSON from aws-crt-cpp, as we have one in aws-c-common already?

I am working on this now! It pretty much maps 1:1 with what we have in `aws-c-common`.

graebm (approved):

PR #460, opened 2023-01-25
Remove CJSON and use aws-c-common JSON instead

TwistedTwigleg:
*Description of changes:*

Removes CJSON from the repository and uses the JSON implementation in `aws-c-common` instead.

*Design Decisions*

100% preserve the Aws::Crt::JsonObject API. This introduced the following difficulties:
1) The existing API allows "invalid" instances to exist. This worked before, because cJSON is very tolerant of passing in NULL. But the aws-c-common API does not tolerate NULL.
    - Decision: add a ton more NULL checks. I did not add ASSERTs or LOGGING because it was just too many distinct points of failure and the code got really bloated when I started adding them.
2) The old API stored a string explaining why an invalid object failed to load. The aws-c-common API just has a failed-to-load error code.
    - Therefore, stop storing a unique string per object. It was difficult to properly deal with this in copy/move operators. 
3) But there's a `GetErrorMessage()` function that returns `const String&`. Note: that's a *reference*.
    - Therefore, we need an actual `String` that we can return a reference to, which outlives the function call. Decided to keep around 2 global singletons, a "Failed to load JSON" String to re-use for all invalid instances, and a blank String to re-use for all valid instances.
__________


source/JsonObject.cpp
```diff
@@ -16,26 +16,24 @@ namespace Aws
...
         }
 
         JsonObject::JsonObject(const String &value) : m_wasParseSuccessful(true)
         {
-            const char *return_parse_end;
-            m_value = cJSON_ParseWithLengthOpts(value.c_str(), value.length(), &return_parse_end, 0);
+            m_value = aws_json_value_new_from_string(ApiAllocator(), aws_byte_cursor_from_c_str(value.c_str()));
 
-            if (m_value == nullptr || cJSON_IsInvalid(m_value) == 1)
+            if (m_value == nullptr || aws_json_value_is_null(m_value) == true)
```
graebm, line 28: I don't think the `aws_json_value_is_null()` check is necessary anymore?

I believe the contract of `aws_json_value_new_from_string()` is that it will return NULL or a valid aws_json_value that is a string

source/JsonObject.cpp
```diff
@@ -16,26 +16,24 @@ namespace Aws
...
                 m_wasParseSuccessful = false;
-                m_errorMessage = "Failed to parse JSON at: ";
-                m_errorMessage += return_parse_end;
+                m_errorMessage = "Failed to parse JSON: " + value;
             }
         }
 
         JsonObject::JsonObject(const JsonObject &value)
-            : m_value(cJSON_Duplicate(value.m_value, 1 /*recurse*/)), m_wasParseSuccessful(value.m_wasParseSuccessful),
+            : m_value(aws_json_value_duplicate(value.m_value)), m_wasParseSuccessful(value.m_wasParseSuccessful),
```
graebm, line 36: `aws_json_value_duplicate()` may return NULL, even if something valid was passed in (at least in theory, not sure if possible in reality)

in which case m_wasParseSuccessful should not match value.m_parseWasSuccessful, and m_errorMessage should not match m_errorMessage

might need to move the logic down into the actual function body

source/JsonObject.cpp
```diff
@@ -16,26 +16,24 @@ namespace Aws
...
+        JsonObject::JsonObject(aws_json_value *value)
+            : m_value(aws_json_value_duplicate(value)), m_wasParseSuccessful(true)
         {
         }
 
         JsonObject::JsonObject(const String &value) : m_wasParseSuccessful(true)
         {
-            const char *return_parse_end;
-            m_value = cJSON_ParseWithLengthOpts(value.c_str(), value.length(), &return_parse_end, 0);
+            m_value = aws_json_value_new_from_string(ApiAllocator(), aws_byte_cursor_from_c_str(value.c_str()));
```
graebm, line 26:
```suggestion
            m_value = aws_json_value_new_from_string(ApiAllocator(), ByteCursorFromString(value));
```

String knows how long it is, use this helper function to avoid the O(N) strlen() under the hood

source/JsonObject.cpp
```diff
@@ -16,26 +16,24 @@ namespace Aws
     {
         JsonObject::JsonObject() : m_wasParseSuccessful(true) { m_value = nullptr; }
```
graebm, line 17: prior art: m_value is NULL, but m_wasParseSuccessful is true. .... that's misleading

unfortunately, this JsonObject class has a concept of "exists but invalid"

but with our C API, there's no such thing as an invalid aws_json_value. Either it's NULL, or it exists and is valid

So ... we need to square that circle. But we can't break this C++ API, right? Are IoT customers using these Json classes directly?

source/JsonObject.cpp
TwistedTwigleg: I checked and it does return `null` on an error, so you are right that we do not need to check that anymore. It relies on `CJSON.Parse` catching the error and returning NULL, but any error path in `CJSON.Parse` does return `null` and not an null/empty CJSON object.

source/JsonObject.cpp
TwistedTwigleg: I am not sure on if anyone is using this API directly or not on the IoT side. I do think we do not want to break the API if we can help it, but on the other hand, I agree that it is misleading and probably should be fixed...

I will investigate and see if I can find usage of this API to determine if we can change it.

graebm: I just got rid of the `m_parseSuccessful` and `m_errorMessage` members. Now copying is much easier.

Now the class has exactly 1 member, and if it's nullptr then the parse was obviously not successful, and there's only 1 possible error messsage which is "Failed to parse JSON"

TwistedTwigleg (commented):
Cannot approve the PR since I opened it, but it looks good to me. Thanks for making these adjustments!

source/Api.cpp
```diff
@@ -49,22 +46,10 @@ namespace Aws
...
-            hooks.malloc_fn = s_cJSONAlloc;
-            hooks.free_fn = s_cJSONFree;
-            cJSON_InitHooks(&hooks);
-        }
-
-        ApiHandle::ApiHandle(Allocator *allocator) noexcept
-            : m_logger(), m_shutdownBehavior(ApiHandleShutdownBehavior::Blocking)
-        {
-            s_initApi(allocator);
+            JsonObject::OnLibraryInit();
```
DmitriyMusatkin, line 49: not something to fix in this pr, but we should think how to make this better. 
hooks in json are global and we are breaking abstraction boundary when code upsteam from us depends on cjson as well. ex. cpp sdk overrides the hooks we set

source/JsonObject.cpp
```diff
@@ -58,10 +43,8 @@ namespace Aws
                 return *this;
             }
 
-            Destroy();
-            m_value = cJSON_Duplicate(other.m_value, 1 /*recurse*/);
-            m_wasParseSuccessful = other.m_wasParseSuccessful;
-            m_errorMessage = other.m_errorMessage;
+            this->~JsonObject();
+            new (this) JsonObject(other);
             return *this;
```
DmitriyMusatkin, line 48:
```suggestion
           if (this != &other)
            {
                this->~JsonObject();
                new (this) JsonObject(other);
            }

            return *this;
```

DmitriyMusatkin (approved):

source/Api.cpp
graebm: maybe something to make the symbols unique...
find/replace in the source code
or fancy interning trickery

PR #462, opened 2023-01-31
[Aws::Crt::Optional] avoid std::aligned_storage

kedartal:
*Description of changes:*

Checks if `std::optional` is available, and if so, alias `Aws::Crt::Optional<T>` to `std::optional<T>`.
Also, fix a missing `#include <type_traits>` when not aliasing, for `std::aligned_storage`.

### DIFF
--- a/include/aws/crt/Optional.h
+++ b/include/aws/crt/Optional.h
@@ -19,14 +19,14 @@ namespace Aws
             Optional() : m_value(nullptr) {}
             Optional(const T &val)
             {
-                new (&m_storage) T(val);
-                m_value = reinterpret_cast<T *>(&m_storage);
+                new (m_storage) T(val);
+                m_value = reinterpret_cast<T *>(m_storage);
             }
 
             Optional(T &&val)
             {
-                new (&m_storage) T(std::forward<T>(val));
-                m_value = reinterpret_cast<T *>(&m_storage);
+                new (m_storage) T(std::forward<T>(val));
+                m_value = reinterpret_cast<T *>(m_storage);
             }
 
             ~Optional()
@@ -45,8 +45,8 @@ namespace Aws
                     return *this;
                 }
 
-                new (&m_storage) T(std::forward<U>(u));
-                m_value = reinterpret_cast<T *>(&m_storage);
+                new (m_storage) T(std::forward<U>(u));
+                m_value = reinterpret_cast<T *>(m_storage);
 
                 return *this;
             }
@@ -55,8 +55,8 @@ namespace Aws
             {
                 if (other.m_value)
                 {
-                    new (&m_storage) T(*other.m_value);
-                    m_value = reinterpret_cast<T *>(&m_storage);
+                    new (m_storage) T(*other.m_value);
+                    m_value = reinterpret_cast<T *>(m_storage);
                 }
                 else
                 {
@@ -68,8 +68,8 @@ namespace Aws
             {
                 if (other.m_value)
                 {
-                    new (&m_storage) T(std::forward<T>(*other.m_value));
-                    m_value = reinterpret_cast<T *>(&m_storage);
+                    new (m_storage) T(std::forward<T>(*other.m_value));
+                    m_value = reinterpret_cast<T *>(m_storage);
                 }
                 else
                 {
@@ -101,8 +101,8 @@ namespace Aws
 
                 if (other.m_value)
                 {
-                    new (&m_storage) T(*other.m_value);
-                    m_value = reinterpret_cast<T *>(&m_storage);
+                    new (m_storage) T(*other.m_value);
+                    m_value = reinterpret_cast<T *>(m_storage);
                 }
 
                 return *this;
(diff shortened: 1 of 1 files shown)


kedartal:
Context: `std::aligned_storage` is deprecated in c++23 (https://github.com/cplusplus/draft/pull/5282). Instead of awkwardly working around it (ignoring deprecated warnings, setting `_LIBCPP_DISABLE_DEPRECATION_WARNINGS`...) or re-implementing `Aws::Crt::Optional`, using the standard implementation seems simpler moving forward.

Please note that `std::optional` is not functionally identical to the `Aws::Crt::Optional` implementation in c++20 and above -- `constexpr` and monadic operations were added. Such functionality can be used conditionally by testing the standard `__cpp_lib_optional` macro (https://en.cppreference.com/w/cpp/feature_test). I couldn't think of a reason why this would break any existing code, but naturally I might be missing something.

graebm:
We can't simply alias the two types like this.

We used to do that, but later realized it didn't work, and removed the aliasing here: https://github.com/awslabs/aws-crt-cpp/pull/231/commits/402d222870ebe73b550ed26dc9a0cc756f252238
> `using Optional = std::optional;` seems like the smart thing to do if your application is c++17. Except we always compile aws-crt-cpp with c++11, so our compiled code uses the ABI of our custom Aws::Crt::Optional class. And if a c++17 application includes that header, it’s like “oh yeah I’m c++17 so the ABI of this thing is std::optional”
> 
> Solution is: always define our own custom Optional class, even if user’s application is c++17. They’re just different classes, each with its own ABI, even if they have the same API

graebm:
Can we get rid of the deprecation warnings by using alignas instead of aligned_storage, as suggested on this [stackoverflow post](https://stackoverflow.com/questions/71828288/why-is-stdaligned-storage-to-be-deprecated-in-c23-and-what-to-use-instead)?

Or we might be able to just remove the alignment stuff. I don't believe this library is putting anything in its custom Optional class that actually has alignment requirements

kedartal:
@graebm I see, though why is `aws-crt-cpp` always compiled with c++11? Anyway yes, `alignas` can work, or alternatively `std::aligned_alloc` (https://en.cppreference.com/w/c/memory/aligned_alloc). Any preferences?
Removing the alignment could break clients that use the class with their own types, but perhaps that's not a priority -- I'd definitely document it :)

kedartal:
@graebm Something like that (just pushed)?

include/aws/crt/Optional.h
```diff
@@ -196,7 +197,7 @@ namespace Aws
             }
 
           private:
-            typename std::aligned_storage<sizeof(T)>::type m_storage;
+            alignas(T) std::array<char, sizeof(T)> m_storage;
```
graebm, line 200: Seems weird to use std::array, when we could use a simple literal array, as suggested in [P1413R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1413r3.pdf#page=5):

```suggestion
            alignas(T) char m_storage[sizeof(T)];
```

graebm:
> ...`alignas` can work, or alternatively `std::aligned_alloc` (https://en.cppreference.com/w/c/memory/aligned_alloc). Any preferences?

`std::aligned_alloc` would perform an unnecessary allocation. Let's use `alignas` for in-place storage.

graebm:
> why is `aws-crt-cpp` always compiled with c++11? 

If the library is used by multiple applications on the system (like if aws-crt-cpp ends up in a package manager, and is distributed precompiled), we want to be sure it's compatible with our minimum spec C++11.

kedartal:
@graebm you're right about `std::aligned_alloc` -- users could always make "fat" types cheap to move, this shouldn't be the responsibility of the optional container.
I don't think there would be any difference between `std::array<char, sizeof(T)>` and `char[ sizeof(T) ]` in this case, let's use whichever you like.
I wasn't aware that `aws-crt-cpp` is meant to guarantee ABI stability, makes sense; perhaps it's worth mentioning that in the documentation as a requirement, and maybe prepare for allowing ABI versioning (e.g, with linker scripts) since coroutines already landed and senders-receivers are just around the corner :)

kedartal:
@graebm please let me know what you think, I believe there are two options:
- make an ABI compatible change: in libc++, libstdc++ and msvc `std::aligned_storage` boils down to a union of `unsigned char[]` and a type that holds the alignment, so that if we made sure `m_storage` is large enough we should be ok (let's make is unsigned just for aesthetics)
https://github.com/llvm/llvm-project/blob/042f01b28928c7acb3ce44e4e6281bf229bebb1c/libcxx/include/__type_traits/aligned_storage.h#L86
https://gcc.gnu.org/onlinedocs/gcc-4.8.2/libstdc++/api/a01609_source.html#l01693
https://github.com/microsoft/STL/blob/f8b8ad6dea901d360d8165cc6719de47ca9d9ca2/stl/inc/type_traits#L1035
- don't dix: add an explicit warning conditional on c++23 and above to make it easier for users to diagnose

graebm (approved):

kedartal:
@graebm ... for the first option, something like `alignas(T) unsigned char m_storage[ std::max( sizeof(T), sizeof(std::max_align_t) ];`

graebm:
> I wasn't aware that `aws-crt-cpp` is meant to guarantee ABI stability

We don't guarantee it right now. But we aspire to someday

PR #469, opened 2023-02-09
MQTT311 extra connection callbacks

TwistedTwigleg:
*Description of changes:*

Adds additional connection callbacks to MQTT311. Specifically, adds `onConnectionSuccess`, `onConnectionFailure`, and `onConnectionClosed`. These callbacks mirror those in V1.

___________


source/mqtt/MqttClient.cpp
```diff
@@ -40,6 +40,27 @@ namespace Aws
                 {
                     connWrapper->OnConnectionResumed(*connWrapper, returnCode, sessionPresent);
                 }
+                if (connWrapper->OnConnectionSuccess)
```
xiazhvera, line 43: Would it be duplicate to OnConnectionResumed? Would we need OnConnectionSuccess for reconnection as well?

xiazhvera (approved):

source/mqtt/MqttClient.cpp
xiazhvera: I think either way works, as long as we documented that it would be fired on reconnection succeed.

TwistedTwigleg: I think it would duplicate OnConnectionResume and be called on every reconnect. That's what each of the SDKs do currently, I think.

Regardless, I agree that it should be documented regardless. I'll check with the V1 SDK and see how it does it (I think it calls it each time, based on a quick look through the V1 SDK Java code, but better to be sure).

TwistedTwigleg: I just checked and confirmed that the V1 SDK, at least for Java, does emit a ConnectionSuccess on every connection (initial and reconnections), ConnectionFailure on every fail (including retries), and ConnectionClose when the client is fully shutdown and destroyed.
This is how this PR and the other SDK PRs work - so it should be good in terms of functionality. I will clarify this in the documentation though for each of the SDK PRs.

TwistedTwigleg:

PR #470, opened 2023-02-13
Updated crypto bindings to include SHA1, added one-shot polymorphic s…

JonathanHenson:
…upport to better support the cpp sdk API.

This can't actually be bound out to the CPP SDK until I add AES implementations to aws-c-cal, but this gets us the hashing algorithms at least.

### DIFF
--- a/include/aws/crt/crypto/HMAC.h
+++ b/include/aws/crt/crypto/HMAC.h
@@ -88,6 +88,12 @@ namespace Aws
                  */
                 bool Digest(ByteBuf &output, size_t truncateTo = 0) noexcept;
 
+                /**
+                 * Returns the size of the digest for this hmac algorithm. If this object is not valid, it will
+                 * return 0 instead.
+                 */
+                size_t DigestSize() const noexcept;
+
               private:
                 HMAC(aws_hmac *hmac) noexcept;
                 HMAC() = delete;
--- a/include/aws/crt/crypto/Hash.h
+++ b/include/aws/crt/crypto/Hash.h
@@ -15,8 +15,9 @@ namespace Aws
     {
         namespace Crypto
         {
-            static const size_t SHA256_DIGEST_SIZE = 32;
-            static const size_t MD5_DIGEST_SIZE = 16;
+            static const size_t SHA1_DIGEST_SIZE = AWS_SHA1_LEN;
+            static const size_t SHA256_DIGEST_SIZE = AWS_SHA256_LEN;
+            static const size_t MD5_DIGEST_SIZE = AWS_MD5_LEN;
 
             /**
              * Computes a SHA256 Hash over input, and writes the digest to output. If truncateTo is non-zero, the digest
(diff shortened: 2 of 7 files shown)


include/aws/crt/crypto/SymmetricCipher.h
```diff
@@ -0,0 +1,147 @@
...
+                /**
+                 * Returns the encryption tag generated during encryption operations for this cipher in GCM mode.
+                 * This tag is not copied from the cipher so do not mutate this
+                 * data. Copy if if you need to pass it around anywhere.
+                 */
+                ByteCursor GetTag() const noexcept;
+
+              private:
+                SymmetricCipher(aws_symmetric_cipher *cipher) noexcept;
+                aws_symmetric_cipher *m_cipher;
```
sbiscigl, line 142: could we make this a unique ptr instead of raw? we could then in the ctor have something like
```
SymmetricCipher::SymmetricCipher(aws_symmetric_cipher *cipher) noexcept : 
  m_cipher(cipher, aws_symmetric_cipher_destroy), m_lastError(0)
{
  ...
}
```

instead of manually have to call it in the dtor. additonally i think you will be then able to use the default move assignment and ctor as well.

source/Types.cpp
```diff
@@ -72,31 +72,59 @@ namespace Aws
...
+                    return outputStr;
+                }
+            }
+
+            return {};
+        }
+
+        namespace UnsafeInteropHelpers
+        {
+            void Base64Decode(const ByteCursor &toDecode, ByteBuf &out) noexcept
```
sbiscigl, line 100: so hopefully i understand this, but i think this is a helper function that can be summarized as "dont attempt to decode if we dont have the space for it, if we do, decode into the buffer". Do we want to expose this as a public API or have a be a impl detail of the the existing API for Base64, leaning towards it as a impl detail of the existing API.

DmitriyMusatkin: i rolled this back for now. im not against having those helpers in general. my major concern is that underlying crt decode impl always starts decoding into output buffer at 0, while the binding assumes it would start appending to output buffer (which it probably should) and we should not expose that foot gun until we fix it. 

second concern is the naming. calling them unsafe, while there are 30 other places where we do the same thing for crypto, seems very inconsistent. maybe we can do something better where we can avoid allocating memory and also avoid unsafeness of byte buf

include/aws/crt/crypto/SymmetricCipher.h
```diff
@@ -0,0 +1,146 @@
...
+                 * Decrypts any remaining data on the cipher and stores the output in out. Out can be dynamically
+                 * re-sized if out is a dynamic byte buf. Otherwise, make sure the size of out is at least 2 blocks
+                 * for CBC, CTR, GCM, and Keywrap modes.
+                 *
+                 * Returns true on success. Call
+                 * LastError() for the reason this call failed.
+                 */
+                bool FinalizeDecryption(ByteBuf &out) noexcept;
+
+                bool Reset() noexcept;
```
SergeyRyabinin, line 117: missing docstring

source/Types.cpp
```diff
@@ -51,53 +51,53 @@ namespace Aws
             return aws_byte_cursor_from_array(array, len);
         }
 
-        Vector<uint8_t> Base64Decode(const String &decode)
+        Vector<uint8_t> Base64Decode(const String &decode) noexcept
```
SergeyRyabinin, line 54: > Exceptions
>Calls to Allocator::allocate may throw. 

https://en.cppreference.com/w/cpp/container/vector/vector

While by default it is built is exceptions disabled, idk what is the point of marking methods as noexcept when the inner code can throw (if exceptions are enabled).
I.e. technically `noexcept` is incorrect here.

source/Types.cpp
```diff
@@ -51,53 +51,53 @@ namespace Aws
...
-
-                if (aws_base64_decode(&toDecode, &tempBuf) == AWS_OP_SUCCESS)
-                {
-                    return output;
-                }
+                return {};
             }
+                
+            Vector<uint8_t> output(allocationSize, 0x00);
+            ByteBuf tempBuf = aws_byte_buf_from_empty_array(output.data(), output.capacity());
```
SergeyRyabinin, line 66: .capacity is not size().

.size() will return allocationSize while capacity may return some value >=allocationSize
(of course typically it will return the size, but still capacity != size).

You are not calling resize / anything else that related to the actual capacity, so .size here is more correct.

source/Types.cpp
SergeyRyabinin: well, on another thought, CRT uses Vector with a custom allocator from CRT that does not throw

source/Types.cpp
```diff
@@ -51,53 +51,53 @@ namespace Aws
...
+
+            size_t allocationSize = 0;
+            if (AWS_OP_SUCCESS != aws_base64_compute_encoded_len(toEncode.len, &allocationSize))
+            {
+                return {};
+            }
 
-            size_t allocation_size = 0;
+            String outputStr(allocationSize, 0x00);
+            auto tempBuf = aws_byte_buf_from_empty_array(outputStr.data(), outputStr.capacity());
```
SergeyRyabinin, line 87: why capacity over size if you are not calling resize later?

sbiscigl (approved):
lgtm pending CI pass

SergeyRyabinin (approved):

PR #475, opened 2023-03-06
Mqtt5Client Clean shutdown

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Added the self-reference share_ptr to Mqtt5Client to avoid the client get released before underlying native client clean up the operations and terminated.
- Updated the callbacks to use raws client instead of shared_ptr
- Added a Close() API to handle client shutdown.
- Improve NewMqtt5Client, so it will return nullptr when client creation failed. 


bin/mqtt5_app/main.cpp
```diff
@@ -581,5 +583,6 @@ int main(int argc, char **argv)
     {
         std::cout << "[ERROR]Failed to start the client " << std::endl;
     }
+    mqtt5Client->Close();
```
sbSteveK, line 586: Maybe we can add a comment that this must be called to clean up for customers.

include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -316,6 +310,20 @@ namespace Aws
                     std::shared_ptr<UnsubscribePacket> unsubscribeOptions,
                     OnUnsubscribeCompletionHandler onUnsubscribeCompletionCallback = NULL) noexcept;
 
+                /*
+                 * Tell the client to release the native client and clean up unhandled the resources
```
sbSteveK, line 314: maybe a typo here?

source/mqtt/Mqtt5Client.cpp
```diff
@@ -362,25 +365,43 @@ namespace Aws
...
+                    m_client = aws_mqtt5_client_new(allocator, &clientOptions);
+                }
             }
 
             Mqtt5Client::~Mqtt5Client()
             {
+                aws_rw_lock_wlock(&m_client_lock);
                 if (m_client != nullptr)
                 {
+                    // Techniquely, this should never happen as the Mqtt5Client will not get released until we call
```
sbSteveK, line 379: trivial: same as above /* */ for consistency?
Technically* typo

include/aws/crt/mqtt/Mqtt5Client.h
TwistedTwigleg: Minor: `/*` -> `/**`

source/mqtt/Mqtt5Client.cpp
```diff
@@ -171,46 +173,45 @@ namespace Aws
...
-                        case aws_mqtt5_packet_type::AWS_MQTT5_PT_NONE:
-                        {
-                            publish = std::make_shared<PublishResult>(error_code);
-                            break;
+                            std::shared_ptr<PubAckPacket> packet = std::make_shared<PubAckPacket>(
+                                *(aws_mqtt5_packet_puback_view *)publshCompletionPacket, callbackData->allocator);
+                            publish = std::make_shared<PublishResult>(std::move(packet));
                         }
-                        default: // Invalid packet type
+                        else // This should never happened.
```
TwistedTwigleg, line 190: Minor: `//` -> `/* ... */`

source/mqtt/Mqtt5Client.cpp
```diff
@@ -171,46 +173,45 @@ namespace Aws
...
-                    if (callbackData->onPublishCompletion != NULL)
+                    case aws_mqtt5_packet_type::AWS_MQTT5_PT_NONE:
                     {
-                        callbackData->onPublishCompletion(callbackData->client, error_code, publish);
+                        publish = std::make_shared<PublishResult>(error_code);
+                        break;
                     }
-
-                    Crt::Delete(callbackData, callbackData->allocator);
+                    default: // Invalid packet type
```
TwistedTwigleg, line 202: Minor: `//` -> `/* ... */`

source/mqtt/Mqtt5Client.cpp
```diff
@@ -362,25 +365,43 @@ namespace Aws
...
+                }
             }
 
             Mqtt5Client::~Mqtt5Client()
             {
+                aws_rw_lock_wlock(&m_client_lock);
                 if (m_client != nullptr)
                 {
+                    // Techniquely, this should never happen as the Mqtt5Client will not get released until we call
+                    // Close(), and in Close() we should have released the native client already. WARN THE USER THERE IS
```
TwistedTwigleg, line 380: Minor: `Close()` is missing "`"

source/mqtt/Mqtt5Client.cpp
```diff
@@ -362,25 +365,43 @@ namespace Aws
...
+                    // SOMETHING WRONG HERE.
+                    AWS_LOGF_WARN(
```
TwistedTwigleg, line 382: Debatable: I'd make this an ERROR

source/mqtt/Mqtt5Client.cpp
```diff
@@ -389,20 +410,59 @@ namespace Aws
                 }
 
                 toSeat = new (toSeat) Mqtt5Client(options, allocator);
-                return std::shared_ptr<Mqtt5Client>(
+                // Creation failed, make sure we release the allocated memory
```
TwistedTwigleg, line 413: Minor: `//` -> `/* ... */`

source/mqtt/Mqtt5Client.cpp
```diff
@@ -389,20 +410,59 @@ namespace Aws
...
             int Mqtt5Client::LastError() const noexcept { return aws_last_error(); }
 
-            bool Mqtt5Client::Start() const noexcept { return aws_mqtt5_client_start(m_client) == AWS_OP_SUCCESS; }
+            bool Mqtt5Client::Start() noexcept
+            {
+                if (aws_rw_lock_try_rlock(&m_client_lock) != AWS_ERROR_SUCCESS || m_client == nullptr)
+                {
+                    AWS_LOGF_ERROR(
+                        AWS_LS_MQTT5_CLIENT,
+                        "Mqtt5Client: Invalid Client or the client is in termination. Please create a new client.");
```
TwistedTwigleg, line 436: Minor: I'd either change `Invalid Client ...` to `Invalid client` or change `... the client is ...` to `... the Client is ...` for consistency. Same change will be needed for `stop()`, `publish()`, etc.

source/mqtt/Mqtt5Client.cpp
```diff
@@ -516,19 +600,40 @@ namespace Aws
                 return result == AWS_OP_SUCCESS;
             }
 
+            void Mqtt5Client::Close() noexcept
+            {
+                aws_rw_lock_wlock(&m_client_lock);
+                if (m_client != nullptr)
+                {
+                    aws_mqtt5_client_release(m_client);
+                    // Waiting for client termination
```
TwistedTwigleg, line 609: Minor: `//` -> `/* ... */`

source/mqtt/Mqtt5Client.cpp
```diff
@@ -171,46 +173,45 @@ namespace Aws
...
-                        case aws_mqtt5_packet_type::AWS_MQTT5_PT_NONE:
-                        {
-                            publish = std::make_shared<PublishResult>(error_code);
-                            break;
+                            std::shared_ptr<PubAckPacket> packet = std::make_shared<PubAckPacket>(
+                                *(aws_mqtt5_packet_puback_view *)publshCompletionPacket, callbackData->allocator);
+                            publish = std::make_shared<PublishResult>(std::move(packet));
                         }
-                        default: // Invalid packet type
+                        else /* This should never happened. */
```
bretambrose, line 190: I'd make it a fatal assert.

source/mqtt/Mqtt5Client.cpp
```diff
@@ -171,46 +173,45 @@ namespace Aws
...
-                        callbackData->onPublishCompletion(callbackData->client, error_code, publish);
+                        publish = std::make_shared<PublishResult>(error_code);
+                        break;
                     }
-
-                    Crt::Delete(callbackData, callbackData->allocator);
+                    default: /* Invalid packet type */
+                    {
+                        AWS_LOGF_INFO(AWS_LS_MQTT5_CLIENT, "Invalid Packet Type.");
+                        publish = std::make_shared<PublishResult>(AWS_ERROR_INVALID_ARGUMENT);
```
bretambrose, line 205: Invalid argument is intended for user-submitted data, not crt-generated data.  I would use a different error code here and above.

source/mqtt/Mqtt5Client.cpp
```diff
@@ -362,25 +365,47 @@ namespace Aws
                 clientOptions.client_termination_handler = &Mqtt5Client::s_clientTerminationCompletion;
                 clientOptions.client_termination_handler_user_data = this;
 
-                m_client = aws_mqtt5_client_new(allocator, &clientOptions);
+                if (aws_rw_lock_init(&m_client_lock) == AWS_OP_SUCCESS)
```
bretambrose, line 368: I would just fatal assert this.

source/mqtt/Mqtt5Client.cpp
```diff
@@ -362,25 +365,47 @@ namespace Aws
...
+                    m_client = aws_mqtt5_client_new(allocator, &clientOptions);
+                }
             }
 
             Mqtt5Client::~Mqtt5Client()
             {
+                aws_rw_lock_wlock(&m_client_lock);
                 if (m_client != nullptr)
                 {
+                    /**
```
bretambrose, line 379: Really skeptical that any of this (if-block) is necessary or useful.

source/mqtt/Mqtt5Client.cpp
```diff
@@ -389,20 +414,59 @@ namespace Aws
...
             }
 
             Mqtt5Client::operator bool() const noexcept { return m_client != nullptr; }
 
             int Mqtt5Client::LastError() const noexcept { return aws_last_error(); }
 
-            bool Mqtt5Client::Start() const noexcept { return aws_mqtt5_client_start(m_client) == AWS_OP_SUCCESS; }
+            bool Mqtt5Client::Start() noexcept
+            {
+                if (aws_rw_lock_try_rlock(&m_client_lock) != AWS_ERROR_SUCCESS || m_client == nullptr)
```
bretambrose, line 436: This pattern doesn't unlock a taken read lock if the client is null.

Maybe we should create a simple scoped rw lock class where creation fatal-assert-takes the lock and destruction releases it.

Also you can't release the lock until the C function returns.  Close() could have decrefed it inbetween unlock and aws_mqtt5_* invocation.

These apply to all usages of this pattern.

source/mqtt/Mqtt5Client.cpp
```diff
@@ -516,19 +604,40 @@ namespace Aws
...
 
+            void Mqtt5Client::Close() noexcept
+            {
+                aws_rw_lock_wlock(&m_client_lock);
+                if (m_client != nullptr)
+                {
+                    aws_mqtt5_client_release(m_client);
+                    /* Waiting for client termination */
+                    std::unique_lock<std::mutex> lock(m_terminationMutex);
+                    m_terminationCondition.wait(lock, [this] { return m_terminationPredicate == true; });
```
bretambrose, line 615: This is a deadlock if someone calls Close() from an event loop triggered callback.

Why do we need to wait anyways?

source/mqtt/Mqtt5Client.cpp
bretambrose: // comments are fine in C++

source/mqtt/Mqtt5Client.cpp
xiazhvera: I checked out the error codes, but I did not find anything suitable for the case here, I used AWS_ERROR_UNKNOWN for now.

xiazhvera:
New solution proposed. Close the PR.

PR #478, opened 2023-03-08
Scoped rwlock

xiazhvera:
*Issue #, if available:*

*Description of changes:*


include/aws/crt/ScopedRWLock.h
```diff
@@ -21,7 +21,6 @@ namespace Aws
           public:
             ScopedTryReadLock() : m_lock(nullptr), m_last_error(AWS_ERROR_INVALID_ARGUMENT) {}
```
bretambrose, line 22: Why even have a default constructor?

include/aws/crt/ScopedRWLock.h
```diff
@@ -0,0 +1,50 @@
...
+    namespace Crt
+    {
+        /**
+         * Custom implementation of an ScopedReadLock type. Wrapping the aws_rw_lock.
+         * On creation, the ScopedTryReadLock will attempts to acquire the lock and returns immediately if it can not.
+         * The lock will be unlocked on destruction.
+         * Use aws_last_error() or operator bool() to check if the lock get acquired successfully.
+         */
+
+        class ScopedTryReadLock
```
bretambrose, line 19: Also, why are we using try and not just lock?

xiazhvera: Try lock will immediately return if it failed to acquire the lock. If we are in termination, then we would like to directly failed the operation instead of waiting on the lock.

xiazhvera: I setup in the way that the write lock will lock once `Close()` is invoked and only get unlock/clean up in destructor. In that case, we can fail all operations after `Close()` invoked.

bretambrose: The write lock would only be held for as long as it take to dec ref the native client and change the pointer.  I don't see why we need to hold it arbitrarily long until the destructor.

xiazhvera: That makes sense. Similarly, I will make a scopedWriteLock for the write lock, so it would unlock once it exit the `Close()` function call. 
Also, I will keep the **Try**ReadLock if no further comments on it.

PR #482, opened 2023-04-06
Add DIY dns resolver for cases where the defaults in CRT are insuffic…

JonathanHenson:
…ient.

### DIFF
--- a/include/aws/crt/io/HostResolver.h
+++ b/include/aws/crt/io/HostResolver.h
@@ -44,6 +44,108 @@ namespace Aws
             };
 
             /**
+             * Interface for providing your own DNS resolver. This allows you to entirely swap out the DNS
+             * implementation for the entire system. To do so, inherit from this class and implement each of the
+             * protected functions. Keep in mind that none of these functions are allowed to block.
+             */
+            class AWS_CRT_CPP_API CustomHostResolverBase : public HostResolver
+            {
+              public:
+                explicit CustomHostResolverBase(Allocator *allocator = ApiAllocator()) noexcept;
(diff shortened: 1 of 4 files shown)


TingDaoK (commented):
My biggest concern about this:
- If customer creates their own impl of `CustomHostResolverBase`, the native source is refcount from C code.
- So, there is a possibility when the C code is keeping the native source, but the C++ object is already end-of-life from C++ perspective. 

Mike had this concern when I added refcount to the input stream, PR is [here](https://github.com/awslabs/aws-crt-cpp/pull/341). And we had a workaround to have a RefCounted class [here](https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/RefCounted.h), so that every time if C code requires on the native pointer, it will actually increase the refcount for the C++ object as well

~Oops, seems like you handled it. NM.~ Edited again... Sorry that I am not familiar with C++ enough.

I did a test
```

TestCustomResolver *testCustomResolver = new TestCustomResolver(allocator);
aws_host_resolver *resolver = testCustomResolver->GetUnderlyingHandle();
aws_host_resolver_acquire(resolver);
/* C++ side has the object released */
delete testCustomResolver;

/* C side still think the resolver is alive as it's acquired before, and try to use it. */
aws_string *hostName = aws_string_new_from_c_str(allocator, "localhost");
aws_host_resolver_resolve_host(resolver, hostName, s_onTestHostResolved, nullptr, nullptr);
```
After that the C++ object functions are cleaned up. Even though the native pointer is still kept alive by the refcount.

JonathanHenson:
Closing as we were able to ship the requested DNS feature before this was needed.

PR #489, opened 2023-04-17
Enable MQTT5 client unit tests in CI and improve MQTT5 client error handling

xiazhvera:
*Issue #, if available:*

*Description of changes:*
1. As the latest c-common change added a way to skip a test. We enable the Mqtt5 Unit tests in the CI and skip it if the environment variable is not set
2. The Mqtt5 client and builder will now return nullptr on failure. 

### DIFF
--- a/tests/Mqtt5ClientTest.cpp
+++ b/tests/Mqtt5ClientTest.cpp
@@ -504,7 +504,7 @@ static int s_TestMqtt5DirectConnectionMinimal(Aws::Crt::Allocator *allocator, vo
     if (!mqtt5TestVars)
     {
         printf("Environment Variables are not set for the test, skip the test");
-        return AWS_ERROR_SUCCESS;
+        return AWS_ERROR_UNKNOWN;
     }
 
     ApiHandle apiHandle(allocator);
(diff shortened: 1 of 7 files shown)


tests/Mqtt5ClientTest.cpp
```diff
@@ -101,11 +101,17 @@ struct Mqtt5TestEnvVars
             {
                 m_error |= aws_get_environment_value(allocator, s_mqtt5_test_envName_direct_hostName, &m_hostname);
                 m_error |= aws_get_environment_value(allocator, s_mqtt5_test_envName_direct_port, &m_port);
-                if (m_error == AWS_OP_SUCCESS)
+                if (m_error != AWS_OP_SUCCESS)
                 {
-                    m_hostname_string = aws_string_c_str(m_hostname);
-                    m_port_value = static_cast<uint16_t>(atoi(aws_string_c_str(m_port)));
+                    return;
```
TwistedTwigleg, line 106: Should these empty `return` statements return `AWS_OP_SKIP`?

tests/Mqtt5ClientTest.cpp
```diff
@@ -101,11 +101,17 @@ struct Mqtt5TestEnvVars
...
+                }
+                if (m_hostname == NULL || m_port == NULL)
+                {
+                    m_error = AWS_OP_ERR;
+                    return;
```
TwistedTwigleg, line 111: Return `m_error`?

TwistedTwigleg (approved):

tests/Mqtt5ClientTest.cpp
xiazhvera: The environment variables are handled in the constructor of `Mqtt5TestEnvVars`, there is nothing to return. The `Mqtt5TestEnvVars` is used to handle the environment vars. We will check the m_error to determine if it successfully process the ENV.

tests/Mqtt5ClientTest.cpp
xiazhvera: Same to the previous case.

PR #492, opened 2023-04-24
Add token key name and value parameters. Add more logging

TwistedTwigleg:
*Description of changes:*

The current MQTT311 custom auth implementation is missing the ability to set the custom authorizer token key name and token key value, making it impossible to login with a signed custom authorizer using the builder.

This PR adds the support for the custom auth token key name and token key value. The change is completely backwards compatible as the new values are optional via a new function definition.

It also adds some more logging around invalid data in the MQTT builder.

____________

### DIFF
--- a/tests/MqttClientTest.cpp
+++ b/tests/MqttClientTest.cpp
@@ -149,7 +149,7 @@ static int s_TestIoTMqtt311ConnectWithNoSigningCustomAuth(Aws::Crt::Allocator *a
     error |= aws_get_environment_value(allocator, s_mqtt5_test_envName_iot_nosign_custom_auth_username, &username);
     error |= aws_get_environment_value(allocator, s_mqtt5_test_envName_iot_nosign_custom_auth_password, &password);
 
-    if (error != AWS_OP_SUCCESS)
+    if (error != AWS_OP_SUCCESS || endpoint == NULL || authname == NULL || username == NULL || password == NULL)
     {
         printf("Environment Variables are not set for the test, skip the test");
         return AWS_OP_SKIP;
@@ -182,6 +182,7 @@ static int s_TestIoTMqtt311ConnectWithNoSigningCustomAuth(Aws::Crt::Allocator *a
     std::promise<void> connectionClosedPromise;
     auto onConnectionCompleted =
         [&](Aws::Crt::Mqtt::MqttConnection &, int errorCode, Aws::Crt::Mqtt::ReturnCode returnCode, bool) {
+            (void)returnCode;
             if (errorCode)
             {
                 connectionCompletedPromise.set_value(false);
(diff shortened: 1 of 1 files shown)


sbSteveK (approved):

TwistedTwigleg:

source/iot/MqttClient.cpp
```diff
@@ -322,10 +355,20 @@ namespace Aws
             {
                 usernameString = AddToUsernameParameter(usernameString, authorizerName, "x-amz-customauthorizer-name=");
             }
-            if (!authorizerSignature.empty())
+            if (!authorizerSignature.empty() || !tokenKeyName.empty() || !tokenValue.empty())
             {
+                if (authorizerSignature.empty() || tokenKeyName.empty() || tokenValue.empty())
```
TingDaoK, line 360: this is a behavior change.

TwistedTwigleg: Yeah, this is a behavior change and is *technically* breaking, but it was never working before. If you set `authorizerSignature` in the old API, you still never could connect to a signed custom authorizer because you do not have the token key name nor token value set. So with the old API, setting `tokenSignature` never worked anyway because of the other two values missing. Prior to this change it would set the token signature in the query parameters, but IoT core always rejects it due to the missing token data.

Now with the new API, you HAVE to set all three or it will fail with an error. This means that if you are using the old API but you have `tokenSignature` set, instead of it "working" but never being able to connect to IoT Core because it doesn't have the right data, it will instead error with an error explaining why.

It is also backwards compatible because the old API still works if you leave `tokenSignature` empty. The old API only worked if `tokenSignature` was empty OR you were connecting to a non-signed custom authorizer, and so that stuff still works, but now for signed authorizers the old API will error out always because it cannot work unless you use the new API.

So technically it's a breaking change but because it never worked before anyway, it's not really breaking because it didn't work, if that makes sense.

TwistedTwigleg: We didn't catch this in our CI nor testing because we didn't test the MQTT311 custom authorizer implementation against a signed custom authorizer - we only tested with unsighed authorizers. Luckily customers didn't find this issue either for C++, but we've had tickets cut for Javascript and C++, both which suffered from this issue (and have the same logic that throws an error unless all three are defined, but skips if none of the three arguments are defined).

TwistedTwigleg: Well, looking at how [Bret did it for Python](https://github.com/aws/aws-iot-device-sdk-python-v2/blob/main/awsiot/mqtt_connection_builder.py#L626), it looks like he did it where it still allows setting the signature without the token values, to be fully backwards compatible. I will make a follow-up PR to adjust it here and across the others.

PR #494, opened 2023-04-24
Full backwards compat with custom auth in MQTT311

TwistedTwigleg:
*Description of changes:*

~~Minor adjustment to stay 100% fully backwards compatible with the MQTT311 custom authorizer API, even if it technically won't/shouldn't work with the old API.~~

Adjusts documentation to note that passing a custom authorizer signature with the old API will always fail.

_________

### DIFF
--- a/source/iot/MqttClient.cpp
+++ b/source/iot/MqttClient.cpp
@@ -355,14 +355,9 @@ namespace Aws
             {
                 usernameString = AddToUsernameParameter(usernameString, authorizerName, "x-amz-customauthorizer-name=");
             }
-            if (!authorizerSignature.empty())
+            if (!authorizerSignature.empty() || !tokenKeyName.empty() || !tokenValue.empty())
             {
-                usernameString =
-                    AddToUsernameParameter(usernameString, authorizerSignature, "x-amz-customauthorizer-signature=");
-            }
-            if (!tokenKeyName.empty() || !tokenValue.empty())
-            {
-                if (tokenKeyName.empty() || tokenValue.empty())
+                if (authorizerSignature.empty() || tokenKeyName.empty() || tokenValue.empty())
                 {
                     AWS_LOGF_ERROR(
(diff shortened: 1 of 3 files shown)


TingDaoK (commented):
If the old API will **always** fail eventually, I'd actually prefer it to fail as the `AWS_ERROR_INVALID_ARGUMENT` raised. 

So, instead of changing how you make it fail, I'd suggest to change the documentation of the old API to be clear that it never works and will not work.

TwistedTwigleg:
Okay, I can do that. I just double checked because I want to make sure I have the most up to date information, and  according to the [custom authorizer documentation](https://docs.aws.amazon.com/iot/latest/developerguide/config-custom-auth.html):

> Token key name: The key name used to extract the token from the HTTP headers, query parameters, or MQTT CONNECT user name in order to perform signature validation. This value is required if signing is enabled in your authorizer. 

So it is required if you use a signed authorizer to validate the signature. I will update the documentation instead then.

source/iot/MqttClient.cpp
```diff
@@ -355,9 +355,24 @@ namespace Aws
...
+            if (!authorizerSignature.empty())
             {
-                if (authorizerSignature.empty() || tokenKeyName.empty() || tokenValue.empty())
+                if (tokenKeyName.empty() || tokenValue.empty())
+                {
+                    AWS_LOGF_WARN(
+                        AWS_LS_MQTT_CLIENT,
+                        "id=%p: Signed custom authorizers with signature will not work without a token key name and "
+                        "token value. Your connection may be rejected/stalled on the IoT Core side due to this. Please "
+                        "use the non-deprecate API and pass both the token key name and token value to connect to a "
```
bretambrose, line 366: deprecated

bretambrose (approved):

waahm7 (approved):

TwistedTwigleg:

PR #495, opened 2023-04-27
Fix null env var handling in Mqtt311ClientTest

lrstewart:
*Description of changes:*

s2n-tls runs the CRT tests on pull requests to avoid breaking CRT. We don't set up these environment variables, so if I understand the intention correctly than these tests should be skipped.

However, skipping IoTMqtt311ConnectWithNoSigningCustomAuth causes a memory leak and IoTMqtt311ConnectWithSigningCustomAuth doesn't skip even if none of the variables are set. This is breaking the s2n-tls PR checks.

I've confirmed locally that the previously failing tests now succeed with this fix.


lrstewart:
Are the checks not running for me because I'm a first time contributor?

TwistedTwigleg:
Thank you for making this pull request! I have added this fix into #494, which also slightly adjusts the custom authorizer behavior.

For security reasons, we do not allow CI to run from external contributors who are not part of specific GitHub teams, so that is why CI didn't run for your PR.

TwistedTwigleg:
Thanks again for this PR! The fix was merged as part of #494. Closing this PR...

PR #499, opened 2023-05-12
Additional unit-tests and unit-test adjustment

TwistedTwigleg:
*Description of changes:*

Adds new unit tests to check for MQTT311 and MQTT5 connections for:
* PKCS11
* PKCS12
* Profile credentials provider
* Environment credentials provider
* Windows Certificate Store
* etc.

It uses a new [CRT builder action here](https://github.com/awslabs/aws-crt-builder/pull/231) to setup the environment variables. Finally, it adjusts Codebuild to also use the CRT builder so everything has the same setup across the board.

**Note:* This PR is always going to be marked as "failing" due to `aws-crt-cpp-integration-tests` in Codebuild, as it uses an older container that doesn't support a new-enough version of Python. The `aws-crt-cpp-integration-tests-tmp` is the equivalent, but with a newer container. I will update `aws-crt-cpp-integration-tests` with the container prior to merge.

________


.github/workflows/ci.yml
```diff
@@ -18,6 +18,25 @@ env:
...
+          - rhel8-x64
+    steps:
+    - name: Install qemu/docker
+      run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+    - name: Build ${{ env.PACKAGE_NAME }}
+      run: |
+        aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
+        ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ matrix.image }} build -p ${{ env.PACKAGE_NAME }} --cmake-extra=-DUSE_OPENSSL=ON
+
+  # The following containers need to run without -DUSE_OPENSSL because they do not have OpenSSL packages
```
sbSteveK, line 37: Trivial: For clarity, we might want to specify "linux-compat" instead of "following containers" because it appears as though "linux-compiler-compat" and onward run using -DUSE_OPENSSL

CMakeLists.txt
```diff
@@ -317,6 +277,19 @@ endif()
 set_target_properties(${PROJECT_NAME} PROPERTIES LINKER_LANGUAGE CXX)
 set_target_properties(${PROJECT_NAME} PROPERTIES CXX_STANDARD ${CMAKE_CXX_STANDARD})
 
+# Skip signing the code - otherwise iOS will not compile in CI at all! Newer versions
```
sbSteveK, line 280: trivial: Reword this? There's some repetition in the description. I've seen both conversational and matter of fact comments throughout our code and don't feel strongly either way but it may be clearer and to the point to remove excess wording? e.g. "This only really applies to iOS currently, as MacOS isn't as picky about it" -> "This only applies to iOS as MacOS doesn't care.". Very nitpicking here -_-;;

tests/CMakeLists.txt
```diff
@@ -20,9 +20,36 @@ if(NOT BYO_CRYPTO)
...
+    add_net_test_case(IoTMqtt311ConnectWithPKCS11)
     add_net_test_case(IoTMqtt311ConnectWithPKCS12)
+    add_net_test_case(IoTMqtt311ConnectWithWindowsCert)
+    add_net_test_case(IoTMqtt311ConnectWSDefault)
+    add_net_test_case(IoTMqtt311ConnectWSStatic)
+    add_net_test_case(IoTMqtt311ConnectWSCognito)
+    add_net_test_case(IoTMqtt311ConnectWSProfile)
+    add_net_test_case(IoTMqtt311ConnectWSEnvironment)
+
+    if(ENABLE_MQTT5_TEST)
```
sbSteveK, line 36: I think we can remove this bool from here since net tests skip automatically if we're not connected to a network and the individual tests skip if it's missing an env variable they need? If we do remove this, we can also remove it from the CMakeList file that sets it (which may mean it's just always on unless something changes it elsewhere)

tests/IotServiceTest.cpp
```diff
@@ -7,26 +7,54 @@
...
+
 static int s_TestIotPublishSubscribe(Aws::Crt::Allocator *allocator, void *ctx)
 {
     using namespace Aws::Crt;
     using namespace Aws::Crt::Io;
     using namespace Aws::Crt::Mqtt;
 
-    const char *credentialFiles[] = {TEST_CERTIFICATE, TEST_PRIVATEKEY, TEST_ROOTCA};
+    // ========== Environment variable setup
+    aws_string *input_host;
```
sbSteveK, line 35: Trivial: The Environment variable setup of aws_strings, envResult, and isEndValid is repeated in multiple tests. Could we set this up as its own singular struct that can be used across all the tests? the isEnvValid could also be set to be false if any of the envResults don't return an AWS_OP_SUCCESS so we don't need to expose an additional int envResult. That would also allow the skip test cleanup to be done in a call to a function with the collection of strings rather than repeating the string destroy block of code. Again, very trivial.

sbSteveK: We could alternately have a function that takes the 4 strings and returns a bool on whether they've all been successfully set or not. On fail they can be cleaned up in the function and you can immediately skip. Or we can just leave it as is since it's perfectly fine and I'm just overthinking this >,<

tests/MqttClientTest.cpp
```diff
@@ -223,242 +155,483 @@ static int s_TestIoTMqtt311ConnectWithNoSigningCustomAuth(Aws::Crt::Allocator *a
...
+    Aws::Crt::Io::SocketOptions socketOptions;
+    socketOptions.SetConnectTimeoutMs(3000);
+    std::shared_ptr<Aws::Crt::Mqtt::MqttConnection> connection = client.NewConnection(
+        aws_string_c_str(endpoint), (uint16_t)std::stoi(aws_string_c_str(port)), socketOptions, tlsContext, false);
+    int connectResult = s_ConnectAndDisconnect(connection);
+    if (connectResult != AWS_OP_SUCCESS)
     {
-        printf("Failed to connect");
-        ASSERT_TRUE(false);
+        throw std::runtime_error("ConnectAndDisconnect failed");
```
xiazhvera, line 305: Is there any reason that we choose to use `runtime_error` instead of the original pattern with `ASSERT`?

tests/Mqtt5ClientCredentialsTest.cpp
```diff
@@ -0,0 +1,1185 @@
...
+    ASSERT_TRUE(builder);
+
+    std::promise<bool> connectionPromise;
+    std::promise<void> stoppedPromise;
+
+    s_setupConnectionLifeCycle(builder, connectionPromise, stoppedPromise);
+
+    std::shared_ptr<Aws::Crt::Mqtt5::Mqtt5Client> mqtt5Client = builder->Build();
+
+    ASSERT_TRUE(mqtt5Client);
```
sbSteveK, line 988: Trivial: This 4 line block of ASSERT_TRUE is repeated 14 times and could potentially be a function that accepts the client and the promise?

There are also a number of tests, including this one, that don't wait on the stoppedPromise after the final ASSERT. Feel like we could probably add those since those same tests do assign them?

sbSteveK (approved):

tests/Mqtt5ClientCredentialsTest.cpp
```diff
@@ -0,0 +1,1185 @@
...
+
+AWS_STATIC_STRING_FROM_LITERAL(s_mqtt5_test_envName_iot_x509_endpoint, "AWS_TEST_MQTT5_IOT_CORE_X509_ENDPOINT");
+AWS_STATIC_STRING_FROM_LITERAL(s_mqtt5_test_envName_iot_x509_role_alias, "AWS_TEST_MQTT5_IOT_CORE_X509_ROLE_ALIAS");
+AWS_STATIC_STRING_FROM_LITERAL(s_mqtt5_test_envName_iot_x509_thing_name, "AWS_TEST_MQTT5_IOT_CORE_X509_THING_NAME");
+AWS_STATIC_STRING_FROM_LITERAL(s_mqtt5_test_envName_iot_x509_cert, "AWS_TEST_MQTT5_IOT_CORE_X509_CERT");
+AWS_STATIC_STRING_FROM_LITERAL(s_mqtt5_test_envName_iot_x509_key, "AWS_TEST_MQTT5_IOT_CORE_X509_KEY");
+AWS_STATIC_STRING_FROM_LITERAL(s_mqtt5_test_envName_iot_x509_ca, "AWS_TEST_MQTT5_IOT_CORE_X509_CA");
+
+// Needed to return "success" instead of skip in Codebuild so it doesn't count as a failure
+AWS_STATIC_STRING_FROM_LITERAL(s_mqtt5_test_envName_codebuild, "CODEBUILD_BUILD_ID");
```
xiazhvera, line 101: Just realized that the `SKIPPED` tests will be treated as failure for ctest, sorry that it caused the issue here. 
The issue is fixed in CMake 3.9, while we are using 3.1 as the minimum requirement cross the SDKs. This is just a note that we could probably remove the check when we bump the minimum support for CMake to 3.9.

tests/MqttClientTest.cpp
TwistedTwigleg: It's because, for some reason I wasn't able to reproduce or quite determine, `ASSERT_TRUE` would be counted as a skip in Codebuild rather than an error. I tried to reproduce it in aws-c-common but was unable to, and then I forgot to look into it further as I moved on to Javascript at the time.

I'm not sure what was causing it though. We can probably go back to ASSERT though - I fixed the issues causing the tests to fail.

CMakeLists.txt
```diff
@@ -317,6 +276,17 @@ endif()
...
+# We need to force skip code signing, or iOS will not comiple in CI due to XCode.
+# This only applies to iOS as MacOS doesn't care.
+if (IOS)
+    set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED "NO")
+    set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "")
+    set(XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "")
+    set(XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED "NO")
+    set_target_properties(${PROJECT_NAME} PROPERTIES
+        MACOSX_BUNDLE_INFO_PLIST ${CMAKE_CURRENT_SOURCE_DIR}/plist.in)
+endif()
```
waahm7, line 288: We only need to set `CMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED` and I have [set it as part of CI ](https://github.com/awslabs/aws-crt-cpp/pull/500/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR235)instead of in cmake so that it doesn’t affect other people building from source.

TwistedTwigleg: Ah, I didn’t think about that. I don’t want to change it for everyone… I’ll adjust it! Thanks for catching that!

PR #511, opened 2023-07-06
Use on_connection_* callbacks from aws-c-mqtt

sfod:
*Description of changes:*

Expose the new on_connection_success and on_connection_failure aws-c-mqtt callbacks.


xiazhvera (approved):
Commented with some trivial issues. The PR Looks good to me!

sbSteveK:
I could be wrong but I think a skipped test might report fail if `aws_string_destroy` isn't called on the env variable strings. Just to be sure, could you set some of the envs and not others and check?

sfod:
@sbSteveK In this PR, I replaced `aws_string` with `String` (which is `std::string` with a custom allocator), so string resources are handled automatically. There is only one `aws_string_destroy` left, and it's called for a correspondent `aws_string` unconditionally.
That said, I still run the tests with a subset of env vars set, and everything worked just fine.

PR #527, opened 2023-08-02
Mqtt5to3 adapter

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Bind out the Mqtt5 to Mqtt3 adapter. 
(Still need update submodules: https://github.com/awslabs/aws-c-mqtt/pull/313)

### DIFF
--- a/source/iot/Mqtt5Client.cpp
+++ b/source/iot/Mqtt5Client.cpp
@@ -616,11 +616,18 @@ namespace Aws
 
             bool useWebsocket = m_websocketConfig.has_value();
 
+            auto tlsContext =
+                Crt::Io::TlsContext(m_tlsConnectionOptions.value(), Crt::Io::TlsMode::CLIENT, m_allocator);
+            if (!tlsContext)
+            {
+                return nullptr;
+            }
+
             auto newConnection = client->NewConnection(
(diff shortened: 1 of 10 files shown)


source/mqtt/MqttClient.cpp
```diff
@@ -303,7 +338,7 @@ namespace Aws
             {
                 auto connection = reinterpret_cast<MqttConnection *>(user_data);
 
-                Allocator *allocator = connection->m_owningClient->allocator;
+                Allocator *allocator = connection->m_allocator;
```
xiazhvera, line 341: `m_owningClient` is a mqtt3 client. As now the `m_owningClient` is not necessary set for the case of Mqtt5, use the m_allocator of connection instead of using the allocator from mqtt client.

source/iot/Mqtt5Client.cpp
```diff
@@ -460,99 +588,57 @@ namespace Aws
                 return nullptr;
             }
 
-            uint16_t port = m_port;
```
xiazhvera, line 463: Move the section to buildClientOption()

xiazhvera:
Updated PR at https://github.com/awslabs/aws-crt-cpp/pull/528. Close this one.

PR #528, opened 2023-08-03
Mqtt5to3 adapter 

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Bind out the Mqtt5 to Mqtt3 adapter.
(Still need update submodules: https://github.com/awslabs/aws-c-mqtt/pull/313)

### DIFF
--- a/include/aws/crt/mqtt/Mqtt5Client.h
+++ b/include/aws/crt/mqtt/Mqtt5Client.h
@@ -319,7 +319,7 @@ namespace Aws
                 const Mqtt5ClientOperationStatistics &GetOperationStatistics() noexcept;
 
                 /**
-                 * Create a new connection object from the client5.
+                 * Create a new MqttConnection object from the Mqtt5Client.
                  *
                  * @return std::shared_ptr<Crt::Mqtt::MqttConnection>
                  */
@@ -339,7 +339,7 @@ namespace Aws
             };
 
             /**
-             * The extra option required to build MqttConnection
+             * The extra options required to build MqttConnection
              */
             class Mqtt5to3AdapterOptions
             {
@@ -367,7 +367,7 @@ namespace Aws
 
                 /**
                  * TLS context for secure socket connections.
-                 * If undefined, then a plaintext connection will be used.
+                 * If undefined, a plaintext connection will be used.
                  */
                 Crt::Optional<Crt::Io::TlsConnectionOptions> m_tlsConnectionOptions;
 
@@ -615,6 +615,12 @@ namespace Aws
                 Mqtt5ClientOptions &operator=(Mqtt5ClientOptions &&) = delete;
 
               private:
+                /*
+                 * Allocate and create a new Mqtt5to3AdapterOptions. This function is internally used by Mqtt5Client to
+                 * support the Mqtt5to3Adapter.
+                 *
+                 * @return Mqtt5to3AdapterOptions
+                 */
                 Mqtt5to3AdapterOptions *NewMqtt5to3AdapterOptions() const noexcept;
 
                 /**
@@ -685,7 +691,7 @@ namespace Aws
 
                 /**
                  * TLS context for secure socket connections.
-                 * If undefined, then a plaintext connection will be used.
+                 * If undefined, a plaintext connection will be used.
                  */
                 Crt::Optional<Crt::Io::TlsConnectionOptions> m_tlsConnectionOptions;
 
--- a/include/aws/crt/mqtt/Mqtt5Client.h
+++ b/include/aws/crt/mqtt/Mqtt5Client.h
@@ -5,6 +5,7 @@
  */
 #include <aws/crt/http/HttpConnection.h>
 #include <aws/crt/mqtt/Mqtt5Types.h>
+#include <aws/crt/mqtt/MqttClient.h>
 
 namespace Aws
 {
@@ -26,6 +27,8 @@ namespace Aws
             class UnsubscribePacket;
             class UnSubAckPacket;
             class Mqtt5ClientCore;
+
+            class Mqtt5to3AdapterOptions;
 
             struct AWS_CRT_CPP_API ReconnectOptions
             {
@@ -315,6 +318,13 @@ namespace Aws
                  */
                 const Mqtt5ClientOperationStatistics &GetOperationStatistics() noexcept;
 
+                /**
+                 * Create a new connection object from the client5.
+                 *
+                 * @return std::shared_ptr<Crt::Mqtt::MqttConnection>
+                 */
+                std::shared_ptr<Crt::Mqtt::MqttConnection> NewConnection() noexcept;
+
                 virtual ~Mqtt5Client();
 
               private:
@@ -324,6 +334,47 @@ namespace Aws
                 std::shared_ptr<Mqtt5ClientCore> m_client_core;
 
                 Mqtt5ClientOperationStatistics m_operationStatistics;
+
+                Mqtt5to3AdapterOptions *m_mqtt5to3AdapterOptions;
+            };
+
+            /**
+             * The extra option required to build MqttConnection
+             */
+            class Mqtt5to3AdapterOptions
+            {
+                friend class Mqtt5ClientOptions;
+                friend class Mqtt5ClientCore;
+
+              public:
+                // Default constructor
+                Mqtt5to3AdapterOptions();
+
+              private:
+                Crt::String m_hostName;
+
+                uint16_t m_port;
+
+                bool m_overwriteWebsocket;
+
+                Crt::Mqtt::OnWebSocketHandshakeIntercept m_webSocketInterceptor;
+
+                /**
+                 * Controls socket properties of the underlying MQTT connections made by the client.  Leave undefined to
+                 * use defaults (no TCP keep alive, 10 second socket timeout).
+                 */
+                Crt::Io::SocketOptions m_socketOptions;
+
+                /**
+                 * TLS context for secure socket connections.
+                 * If undefined, then a plaintext connection will be used.
+                 */
+                Crt::Optional<Crt::Io::TlsConnectionOptions> m_tlsConnectionOptions;
+
+                /**
+                 * Configures (tunneling) HTTP proxy usage when establishing MQTT connections
+                 */
+                Crt::Optional<Crt::Http::HttpClientConnectionProxyOptions> m_proxyOptions;
             };
 
             /**
@@ -564,6 +615,8 @@ namespace Aws
                 Mqtt5ClientOptions &operator=(Mqtt5ClientOptions &&) = delete;
 
               private:
+                Mqtt5to3AdapterOptions *NewMqtt5to3AdapterOptions() const noexcept;
+
                 /**
                  * This callback allows a custom transformation of the HTTP request that acts as the websocket
                  * handshake. Websockets will be used if this is set to a valid transformation callback.  To use
--- a/include/aws/crt/mqtt/Mqtt5Client.h
+++ b/include/aws/crt/mqtt/Mqtt5Client.h
@@ -335,11 +335,11 @@ namespace Aws
 
                 Mqtt5ClientOperationStatistics m_operationStatistics;
 
-                Mqtt5to3AdapterOptions *m_mqtt5to3AdapterOptions;
+                ScopedResource<Mqtt5to3AdapterOptions> m_mqtt5to3AdapterOptions;
             };
 
             /**
-             * The extra options required to build MqttConnection
+             * The extra options required to build MqttConnection from Mqtt5Client
              */
             class Mqtt5to3AdapterOptions
             {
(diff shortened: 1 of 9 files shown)


include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -315,6 +318,13 @@ namespace Aws
                  */
                 const Mqtt5ClientOperationStatistics &GetOperationStatistics() noexcept;
 
+                /**
+                 * Create a new connection object from the client5.
+                 *
+                 * @return std::shared_ptr<Crt::Mqtt::MqttConnection>
+                 */
+                std::shared_ptr<Crt::Mqtt::MqttConnection> NewConnection() noexcept;
```
sbSteveK, line 326: Note: We should change the description for MqttConnection as the memory is not going to always be owned by MqttClient anymore.

include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -315,6 +318,13 @@ namespace Aws
                  */
                 const Mqtt5ClientOperationStatistics &GetOperationStatistics() noexcept;
 
+                /**
+                 * Create a new connection object from the client5.
```
sbSteveK, line 322: Trivial: `Mqtt5Client`

sbSteveK: Maybe specify 'MqttConnection` object as well.

include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -324,6 +334,47 @@ namespace Aws
                 std::shared_ptr<Mqtt5ClientCore> m_client_core;
 
                 Mqtt5ClientOperationStatistics m_operationStatistics;
+
+                Mqtt5to3AdapterOptions *m_mqtt5to3AdapterOptions;
```
sbSteveK, line 338: Might be clarified as I get further in but just wondering if we'll only ever have one of these adapter options per Mqtt5Client. Thinking of a scenario where we have a number of different adapters from a single Mqtt5 client. e.g. if someone makes a bunch to feed to different service clients instead of using the same one for them all. May be a non-issue.

include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -324,6 +334,47 @@ namespace Aws
...
+            };
+
+            /**
+             * The extra option required to build MqttConnection
```
sbSteveK, line 342: Trivial: "options"

include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -324,6 +334,47 @@ namespace Aws
...
+
+                /**
+                 * Controls socket properties of the underlying MQTT connections made by the client.  Leave undefined to
+                 * use defaults (no TCP keep alive, 10 second socket timeout).
+                 */
+                Crt::Io::SocketOptions m_socketOptions;
+
+                /**
+                 * TLS context for secure socket connections.
+                 * If undefined, then a plaintext connection will be used.
```
sbSteveK, line 370: Trivial: "If undefined, a plaintext connection will be used."

include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -564,6 +615,8 @@ namespace Aws
                 Mqtt5ClientOptions &operator=(Mqtt5ClientOptions &&) = delete;
 
               private:
+                Mqtt5to3AdapterOptions *NewMqtt5to3AdapterOptions() const noexcept;
```
sbSteveK, line 618: Trivial: Missing description

include/aws/crt/mqtt/Mqtt5ClientCore.h
```diff
@@ -107,6 +107,18 @@ namespace Aws
               private:
                 Mqtt5ClientCore(const Mqtt5ClientOptions &options, Allocator *allocator = ApiAllocator()) noexcept;
 
+                /**
+                 * Create a new connection object over plain text from the client5. The client must outlive
```
sbSteveK, line 111: Trivial `Mqtt5Client`

include/aws/crt/mqtt/Mqtt5ClientCore.h
```diff
@@ -107,6 +107,18 @@ namespace Aws
...
+                 * all of its connection instances.The Mqtt5 Options will be overwritten by the options,
```
sbSteveK, line 112: Trivial: Missing space between period and "The"

sbSteveK: Trivial: unecessary comma at the end of this sentence.

include/aws/crt/mqtt/MqttClient.h
```diff
@@ -460,6 +468,23 @@ namespace Aws
                     const Io::SocketOptions &socketOptions,
                     bool useWebsocket) noexcept;
 
+                MqttConnection(
+                    aws_mqtt5_client *client,
```
sbSteveK, line 472: Should we rename this to mqtt5Client so that we don't accidentally confuse it with the `aws_mqtt_client` that's being used in the other constructors?

include/aws/crt/mqtt/MqttClient.h
```diff
@@ -460,6 +468,23 @@ namespace Aws
...
+                    const char *hostName,
+                    uint16_t port,
+                    const Io::SocketOptions &socketOptions,
+                    const Crt::Io::TlsConnectionOptions &tlsConnectionOptions,
+                    bool useWebsocket,
+                    aws_allocator *allocaotr) noexcept;
+
+                MqttConnection(
+                    aws_mqtt5_client *client,
```
sbSteveK, line 481: rename to mqtt5Client ?

include/aws/crt/mqtt/Mqtt5Client.h
xiazhvera: We only have one adapter options for each Mqtt5Client. When we create MqttConnection from the Mqtt5Client, we will pass the options in the MqttConnection, which will copy the options over.

source/mqtt/Mqtt5Client.cpp
```diff
@@ -233,6 +247,36 @@ namespace Aws
 
             Mqtt5ClientOptions::~Mqtt5ClientOptions() {}
 
+            Mqtt5to3AdapterOptions *Mqtt5ClientOptions::NewMqtt5to3AdapterOptions() const noexcept
+            {
+                Mqtt5to3AdapterOptions *adapterOptions = new Mqtt5to3AdapterOptions();
```
sfod, line 252: Debatable: If `m_mqtt5to3AdapterOptions` is populated directly in this method, then there will be no need in allocating `m_mqtt5to3AdapterOptions` on the heap or even making it a pointer.
My main concern is a `new` call in the code: there is no issue with it, but it always enforces you to make sure that corresponding `delete` will be called. Maybe at least replace it with `std::unique_ptr`?

source/mqtt/Mqtt5Client.cpp
```diff
@@ -233,6 +247,36 @@ namespace Aws
...
+                    adapterOptions->m_tlsConnectionOptions = m_tlsConnectionOptions.value();
+                }
+                if (websocketHandshakeTransform)
+                {
+                    adapterOptions->m_overwriteWebsocket = true;
+
+                    auto signerTransform = [this](
+                                               std::shared_ptr<Crt::Http::HttpRequest> req,
+                                               const Crt::Mqtt::OnWebSocketHandshakeInterceptComplete &onComplete) {
+                        this->websocketHandshakeTransform(req, onComplete);
```
sfod, line 269: `websocketHandshakeTransform` takes its first parameter by value, so `req` should be moved here

include/aws/crt/mqtt/MqttClient.h
```diff
@@ -527,6 +552,12 @@ namespace Aws
                     const char *hostName,
                     uint16_t port,
                     const Io::SocketOptions &socketOptions);
+                static void s_connectionInit(
+                    MqttConnection *self,
+                    const char *hostName,
+                    uint16_t port,
+                    const Io::SocketOptions &socketOptions,
+                    aws_mqtt5_client *mqtt5client);
```
sfod, line 560: Trivial: A small addition to Steve's comment about naming `aws_mqtt5_client` parameters: `mqtt5client` here probably should be renamed to `mqtt5CLient`, just for consistency

sbSteveK:
Trivial: The comment in Mqtt5ClientCore.cpp currently on line 524 `/* Copied from MqttClient.cpp:ln754 (MqttClient::NewConnection) */` Doesn't point to the proper place anymore due to changes in MqttClient.cpp. Let's remove the line reference in that comment.

source/mqtt/Mqtt5Client.cpp
sbSteveK: In the crt we always pair `*_destroy()` for anything that has a `*_new()` that takes an allocator to make sure we clean up anything we allocate. Unsure if we have the same for something with `New` in cpp but I feel like I've seen `New` used that returns a pointer or a null with no corresponding `delete` or `destroy`. Good to stay consistent though so worth investigating.

source/mqtt/Mqtt5Client.cpp
```diff
@@ -21,10 +21,23 @@ namespace Aws
...
             {
                 m_client_core = Mqtt5ClientCore::NewMqtt5ClientCore(options, allocator);
+                m_mqtt5to3AdapterOptions = options.NewMqtt5to3AdapterOptions();
+            }
+
+            std::shared_ptr<Crt::Mqtt::MqttConnection> Mqtt5Client::NewConnection() noexcept
+            {
+                if (m_client_core == nullptr)
+                {
+                    AWS_LOGF_DEBUG(AWS_LS_MQTT5_CLIENT, "Failed to create mqtt3 connection: Mqtt5 Client is invalid.");
```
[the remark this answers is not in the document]
xiazhvera: I think it is a good idea to set the error level to ERROR for those "invalid mqtt5 client" issue here. As if the issue happened (the client core is `nullptr`), it usually means the Mqtt5Client is unusable, and we should expose it to user. 
Since we've been using DEBUG log for all other operations, I will keep it here for now and create a separate PR to update the log level all together.

include/aws/crt/mqtt/MqttClient.h
```diff
@@ -527,6 +552,12 @@ namespace Aws
                     const char *hostName,
                     uint16_t port,
                     const Io::SocketOptions &socketOptions);
+                static void s_connectionInit(
```
sbSteveK, line 555: Super Trivial: empty line between ln554 and ln555

tests/Mqtt5ClientTest.cpp
```diff
@@ -2530,4 +2530,333 @@ static int s_TestMqtt5OperationStatisticsSimple(Aws::Crt::Allocator *allocator,
...
+    std::shared_ptr<Aws::Crt::Mqtt::MqttConnection> mqttConnection = mqtt5Client->NewConnection();
+    ASSERT_TRUE(mqttConnection);
+    int connectResult = s_ConnectAndDisconnect(mqttConnection);
+    ASSERT_SUCCESS(connectResult);
+
+    return AWS_OP_SUCCESS;
+}
+AWS_TEST_CASE(Mqtt5to3AdapterDirectConnectionWithMutualTLS, s_TestMqtt5to3AdapterDirectConnectionWithMutualTLS)
+
+#endif // !BYO_CRYPTO∏
```
sfod, line 2862: Trivial: some strange symbol appeared at the end of line

source/mqtt/MqttClient.cpp
```diff
@@ -295,6 +295,41 @@ namespace Aws
...
+                aws_mqtt5_client *mqtt5client)
+            {
+
+                self->m_hostName = String(hostName);
+                self->m_port = port;
+                self->m_socketOptions = socketOptions;
+
+                self->m_underlyingConnection = aws_mqtt_client_connection_new_from_mqtt5_client(mqtt5client);
+
+                if (self->m_underlyingConnection)
```
sbSteveK, line 312: This section seems identical to the one for the function being overloaded. Is it worth it to split this off into a small helper function that both `s_connectionInit` call? I don't think we even need to pass the function an argument since they'd be using `self->m_underlyingConnection`. Also, there is a way for `self->m_underlyingConnection` to be null so should we at least log an error or something in that case? Currently both do nothing at this point if it's null.

source/mqtt/MqttClient.cpp
```diff
@@ -340,11 +375,39 @@ namespace Aws
...
                 bool useWebsocket) noexcept
-                : m_owningClient(client), m_onAnyCbData(nullptr), m_useTls(false), m_useWebsocket(useWebsocket)
+                : m_owningClient(client), m_onAnyCbData(nullptr), m_useTls(false), m_useWebsocket(useWebsocket),
+                  m_allocator(client->allocator)
             {
                 s_connectionInit(this, hostName, port, socketOptions);
             }
 
+            MqttConnection::MqttConnection(
+                aws_mqtt5_client *client,
```
sbSteveK, line 385: Trivial: `client` -> `mqtt5Client`

source/mqtt/MqttClient.cpp
```diff
@@ -340,11 +375,39 @@ namespace Aws
...
+                bool useWebsocket,
+                aws_allocator *allocator) noexcept
+                : m_owningClient(nullptr), m_tlsOptions(tlsOptions), m_onAnyCbData(nullptr), m_useTls(true),
+                  m_useWebsocket(useWebsocket), m_allocator(allocator)
+            {
+                s_connectionInit(this, hostName, port, socketOptions, client);
+            }
+
+            MqttConnection::MqttConnection(
+                aws_mqtt5_client *client,
```
sbSteveK, line 399: Trivial: `client` -> `mqtt5Client`

sbSteveK (approved):
Looks good. May want to wait for Igor to also check over changes tho.

sfod (approved):

sfod:
Looks good to me too

PR #535, opened 2023-08-29
MQTT3 clean shutdown

sfod:
*Description of changes:*

Add MqttConnectionCore: an auxiliary class handling lifetime of the underlying C connection.

Additional changes:
- Extract MqttConnection class from MqttClient.cpp into separate source files.
- Add MqttConnectionOptions class to pass all possible options on MqttConnection object creation uniformly.
- Add MqttTypes.h with MQTT3-related types used in multiple places.


sfod:
To reviewers:
1. I set a destination branch to `add-sanitizers` to check that data race issue caused by MqttConnection are fixed. After https://github.com/awslabs/aws-crt-cpp/pull/522 is merged, I'll return the `main` branch as the destination.
2. Most changes are caused by moving logic from MqttConnection to MqttConnectionCore, I tried to leave it as is. It doesn't mean that you should ignore most changes in MqttConnectionCore, just keep it in mind.
3. The `CI/clang-sanitizer` job is failing because there are multithreading issues in `elasticurl` (maybe something else too). They'll be fixed in other PRs.

source/mqtt/MqttConnectionCore.cpp
```diff
@@ -0,0 +1,891 @@
...
+                }
+                else
+                {
+                    AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "Failed to initialize Mqtt Connection");
+                }
+            }
+
+            std::shared_ptr<MqttConnection> MqttConnectionCore::obtainConnectionInstance()
+            {
+                // FIXME Okay, this actually doesn't work. There is a chance that the MqttConnection dtor started
```
bretambrose: This sounds alarming.  It seems like we're doing things pretty differently than the 5 client and I don't see why.  I'm not sure I see a path to correctness with this approach.

sfod: I checked once again, and the case I described in this comment is not possible after all. If the `MqttConnection` destructor started its execution, that means that the counter in `std::shared_ptr` reached 0. So, `std::weak_ptr::lock()` won't return object.

source/mqtt/MqttTypes.cpp
```diff
@@ -0,0 +1 @@
+#include <aws/crt/mqtt/MqttTypes.h>
```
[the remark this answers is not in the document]
sfod: Ah, I tested that all includes are in place. I'll remove this file.

include/aws/crt/mqtt/private/MqttConnectionCore.h
```diff
@@ -0,0 +1,383 @@
...
+             * The MqttConnectionCore is an internal class for MqttConnection. The class is used to handle communication
+             * between MqttConnection and underlying C MQTT client. This class should only be used internally by
+             * MqttConnection.
+             * @note Unless otherwise specified, all function arguments need only to live through the duration of the
+             * function call.
+             */
+            class AWS_CRT_CPP_API MqttConnectionCore final : public std::enable_shared_from_this<MqttConnectionCore>
+            {
+              public:
+                /**
```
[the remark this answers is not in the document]
sfod: The intention was to get rid of the private constructors which make impossible to use `make_shared` and force adding friends for the whole class.
If you think it's overcomplicated, it's really easy to revert back to private constructors.

source/mqtt/MqttConnectionCore.cpp
```diff
@@ -0,0 +1,891 @@
...
+                return aws_mqtt_client_connection_disconnect(
+                           m_underlyingConnection, MqttConnectionCore::s_onDisconnect, this) == AWS_OP_SUCCESS;
+            }
+
+            aws_mqtt_client_connection *MqttConnectionCore::GetUnderlyingConnection() noexcept
+            {
+                return m_underlyingConnection;
+            }
+
+            // TODO Fix memory leak on second call.
```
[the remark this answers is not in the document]
sfod: I didn't want to, but yeah, why not :)

source/mqtt/MqttConnectionCore.cpp
bretambrose: Ok, I think I follow now.  I am still skeptical of m_isConnectionAlive though.

include/aws/crt/mqtt/private/MqttConnectionCore.h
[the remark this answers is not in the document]
bretambrose: Feels a bit weird.  Given the existing pattern, let's just stick with it since there doesn't seem to be a clear benefit.  In general, we can definitely break with past patterns when improvements are clear, but this leaves me in kind of a maybe?/maybe not? state of mind.

source/mqtt/MqttConnectionCore.cpp
sfod: Actually, with the current implementation, `m_isConnectionAlive` is not required anymore: `std::weak_ptr` handles the liveliness of the `MqttConnection` object. Moreover, the mutex is not needed anymore as well.

sfod: I added a comment describing how it works.

include/aws/crt/mqtt/MqttTypes.h
```diff
@@ -0,0 +1,132 @@
...
+        namespace Mqtt
+        {
+            class MqttConnection;
+
+            /**
+             * Options required to create an MqttConnection.
+             */
+            struct MqttConnectionOptions
+            {
+                aws_mqtt_client *client = nullptr;
```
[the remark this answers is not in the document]
sfod: Since having `client` and `mqtt5Client` members in this struct may be confusing, I removed them for good.

source/mqtt/MqttConnectionCore.cpp
```diff
@@ -0,0 +1,920 @@
...
+                {
+                    Crt::Delete(opCompleteCallbackData, m_allocator);
+                }
+
+                return packetId;
+            }
+
+            const MqttConnectionOperationStatistics &MqttConnectionCore::GetOperationStatistics() noexcept
+            {
+                aws_mqtt_connection_operation_statistics m_operationStatisticsNative = {0, 0, 0, 0};
```
bretambrose, line 905: locals shouldn't have a m_ prefix for clarity.

bretambrose (approved):

PR #536, opened 2023-08-30
Fix failing CI tests

sfod:


tests/CMakeLists.txt
```diff
@@ -210,7 +210,6 @@ if(NOT BYO_CRYPTO)
     add_net_test_case(Mqtt5NullPublish)
     add_net_test_case(Mqtt5NullSubscribe)
     add_net_test_case(Mqtt5NullUnsubscribe)
-    add_net_test_case(Mqtt5NullConnectPacket)
```
xiazhvera, line 213: Why would we directly remove the test?

sfod: This test checked that it's not possible to create an Mqtt5Client object with the invalid client ID. Since the validation was removed from the underlying lib, this test doesn't make much sense anymore.
I just realized, does this test actually check that if we pass **any** invalid option on a client creation, the creation should fail?

xiazhvera: Yes, the this is an validation test, even if the client id is no longer a case here, we probably should update the test to use another option to make a "invalid connect packet" test. 
Also, looking into the aws-c-mqtt, the client id validation is still there. I think we only updated the validation for mqtt3 so far. If the test failed, it is possible that there is something else went wrong...
https://github.com/awslabs/aws-c-mqtt/blob/ee06ce3db902d1760874c6d950e309cd936f8523/source/v5/mqtt5_options_storage.c#L329

sfod: For history: we decided offline that this function could be safely removed

xiazhvera (approved):

PR #542, opened 2023-09-15
Add Mqtt5ClientBuilder method to set bootstrap

gqmelo:
*Issue #, if available:*

*Description of changes:*

Found this inconsistency when trying to build a Mqtt5Client.
Setting the bootstrap is available when using Mqtt5ClientOptions directly but it was missing from Mqtt5ClientBuilder.

-----------------------

### DIFF
--- a/source/iot/Mqtt5Client.cpp
+++ b/source/iot/Mqtt5Client.cpp
@@ -313,6 +313,12 @@ namespace Aws
             return *this;
         }
 
+        Mqtt5ClientBuilder &Mqtt5ClientBuilder::WithBootstrap(Crt::Io::ClientBootstrap *bootStrap) noexcept
+        {
+            m_options->WithBootstrap(bootStrap);
+            return *this;
+        }
+
         Mqtt5ClientBuilder &Mqtt5ClientBuilder::WithCertificateAuthority(const char *caPath) noexcept
         {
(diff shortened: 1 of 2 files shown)


source/iot/Mqtt5Client.cpp
```diff
@@ -313,6 +313,12 @@ namespace Aws
             return *this;
         }
 
+        Mqtt5ClientBuilder &Mqtt5ClientBuilder::WithBootstrap(Crt::Io::ClientBootstrap *bootStrap) noexcept
+        {
+            m_bootstrap = bootStrap;
```
xiazhvera, line 318: As the builder does not do any extra operations on bootstrap, we would directly pass along the bootstrap to the options. The member variable `m_bootstrap` would not be necessary either. 
```
m_options->WithBootstrap(bootStrap);
```

xiazhvera (changes requested):

source/iot/Mqtt5Client.cpp
gqmelo: oh, that is true, I just saw the member and didn't think about it. Actually the member wasn't even used anywhere so it was not even using the bootstrap. I fixed it now, thanks!

xiazhvera (approved):
Thanks for the contribution!

PR #545, opened 2023-09-27
Adapter API Update

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Move the adapter API into MqttConnection instead of leaving it in Mqtt5Client so that the mqtt5 user would not prompted to use adapter for 5 client.
- Move mqtt5 client core into private header


source/mqtt/MqttConnection.cpp
```diff
@@ -22,6 +24,22 @@ namespace Aws
...
+                    return nullptr;
+                }
+
+                /* Creating a MqttConnection would need access the underlying c mqtt5 client. The c mqtt5 client
+                ** should only be managed by the Mqtt5ClientCore to make sure it is not terminated. Use the client
+                ** core to create a new connection.
+                */
+                return mqtt5client->m_client_core->NewConnection(mqtt5client->m_mqtt5to3AdapterOptions.get());
+            }
+
```
sfod, line 42: The creation process became unnecessary complicated: `MqttConnection` accesses a private member of `Mqtt5Client`, then it calls a private method of `Mqtt5ClientCore`, then that private method calls a private method of `MqttConnection`. So, basically the chain of accesses/calls looks like the following:
```
MqttConnection::createConn -> Mqtt5Client::core -> Mqtt5ClientCore::createConn -> MqttConnection::actualCreateConn
```

All the `MqttConnection::NewConnectionFromMqtt5Client` needs from `Mqtt5Client` is  an `MqttConnectionOptions` object. So one possible way is to add a method `Mqtt5Client::getMqttConnectionOptions`, private or public. It'll make the flow easier to understand.

sfod: The underlying `aws_mqtt5_client` is also required. So, `Mqtt5Client` should expose it to `MqttConnection` too. But still, I think it's better than `MqttConnection` and `Mqtt5Client/Core` being friends to each other and using each other's private members.

xiazhvera: I was worried that if `MqttConnection` could directly access the `aws_mqtt5_client`, we would need make sure `MqttConnection` acquire a reference of `aws_mqtt5_client` so that the c client would not be released during the creation process. If that is the case, I think it would be cleaner that we always have the `Mqtt5ClientCore` to do any operations on `aws_mqtt5_client` as it already acquire a reference of the c client. 
However, thinking of it, if we passed the Mqtt5Client shared_ptr by value here, it should keep the Mqtt5Client alive and keep `Mqtt5ClientCore` and the `aws_mqtt5_client` reference as well. 
It should be good to directly access the `aws_mqtt5_client`, feel free to correct me if I was wrong. 
I will update the code.

source/mqtt/Mqtt5ClientCore.cpp
```diff
@@ -642,6 +609,43 @@ namespace Aws
                 }
             }
 
+            Mqtt5to3AdapterOptions::Mqtt5to3AdapterOptions() {}
+
+            ScopedResource<Mqtt5to3AdapterOptions> Mqtt5to3AdapterOptions::NewMqtt5to3AdapterOptions(
```
xiazhvera, line 614: Debatable: I made the `Mqtt5to3AdapterOptions` as a `ScopedResource`(unique_ptr) so that we could use the aws_allocator to manage the memory of the options. However, I'm also worried it might be over complicated for the case here.

xiazhvera:
Updated and clean up the `NewConnectionFromMqtt5Client` function. Also I moved all adapter related functions to `Mqtt5to3AdapterOptions` in private header.

source/mqtt/Mqtt5ClientCore.cpp
```diff
@@ -642,6 +609,43 @@ namespace Aws
...
+                const Mqtt5ClientOptions &options) noexcept
+            {
+                Allocator *allocator = options.m_allocator;
+                ScopedResource<Mqtt5to3AdapterOptions> adapterOptions = ScopedResource<Mqtt5to3AdapterOptions>(
+                    Crt::New<Mqtt5to3AdapterOptions>(allocator),
+                    [allocator](Mqtt5to3AdapterOptions *options) { Crt::Delete(options, allocator); });
+                adapterOptions->m_mqtt3options.allocator = options.m_allocator;
+                adapterOptions->m_mqtt3options.hostName = options.m_hostName.c_str();
```
sfod, line 622: Storing a pointer to the underlying data of `m_hostName` can cause a use-after-free issue. This code is moved from `Mqtt5ClientCore::NewConnection` as is, but I think it should be addressed in this PR.
Looking at the other members, it seems the issue is with `m_hostName` only, but it's better to check them as well.

source/mqtt/Mqtt5ClientCore.cpp
```diff
@@ -642,6 +609,43 @@ namespace Aws
...
+                }
+                if (options.websocketHandshakeTransform)
+                {
+                    adapterOptions->m_mqtt3options.useWebsocket = true;
+
+                    auto signerTransform = [&options](
+                                               std::shared_ptr<Crt::Http::HttpRequest> req,
+                                               const Crt::Mqtt::OnWebSocketHandshakeInterceptComplete &onComplete) {
+                        options.websocketHandshakeTransform(std::move(req), onComplete);
+                    };
```
sfod, line 640: Same as above: we use `options` by its address, but the `Mqtt5Client` class description doesn't indicate that users have to keep `options` alive after the client creation.

sfod (approved):

PR #553, opened 2023-10-09
Unit test for removing callback severance on 5-to-3 adapter 

xiazhvera:
*Issue #, if available:*

*Description of changes:*


tests/Mqtt5ClientTest.cpp
```diff
@@ -3166,6 +3166,50 @@ static int s_TestMqtt5to3AdapterOperations(Aws::Crt::Allocator *allocator, void
...
+        Mqtt::MqttConnection::NewConnectionFromMqtt5Client(mqtt5Client);
+    ASSERT_TRUE(mqttConnection);
+
+    // Publish an offline message to create an incomplete publish operation
+    mqttConnection->Publish(testTopic.c_str(), Mqtt::QOS::AWS_MQTT_QOS_AT_LEAST_ONCE, false, testPayload, NULL);
+
+    delete builder;
+
+    // If the incomplete operation callback was not called, there would be a memory leak as the callback data was not
+    // released
```
sfod, line 3208: Debatable: Maybe pass to Publish operation a callback which will set flag, something like `is_callback_called`. Then if callback is not called, we have a clear ASSERT statement indicating what the exact problem is. Currently, you have to go to this code and read a comment to figure out why a memory leak appeared.

sfod (approved):

tests/Mqtt5ClientTest.cpp
xiazhvera: Thinking about it, we actually could not find a good timing to validate the `is_callback_called` flag, because we do not know when does the callback is called during the termination process.

PR #575, opened 2023-11-28
Mqtt5 GA API Review

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Verified in https://github.com/aws/aws-iot-device-sdk-cpp-v2/pull/659 that the change would not break current cpp v2. 


sfod:
There are some breaking changes (e.g., `WithRetain()` function removed or new enums being ambiguous). Maybe it'll be better to minimize breaking changes as much as possible?

bin/mqtt5_canary/main.cpp
```diff
@@ -450,7 +450,7 @@ static int s_AwsMqtt5CanaryOperationUnsubscribeBad(struct AwsMqtt5CanaryTestClie
             unsubscription, [testClient](int, std::shared_ptr<Mqtt5::UnSubAckPacket> packet) {
                 if (packet == nullptr)
                     return;
-                if (packet->getReasonCodes()[0] == AWS_MQTT5_UARC_SUCCESS)
+                if (packet->getReasonCodes()[0] == UnSubAckReasonCode::AWS_MQTT5_UARC_SUCCESS)
```
bretambrose, line 453: This isn't what I had in mind for the enums.  Did the original proposal not work?

xiazhvera: As I'm creating the enum using the same value name in aws-c-mqtt (`UnSubAckReasonCode:: AWS_MQTT5_UARC_SUCCESS` v.s. `aws_mqtt5_unsuback_reason_code::AWS_MQTT5_UARC_SUCCESS`), the compiler would report an ambiguity error. I had to add the namespace to avoid the error.

One way to fix it is using a different enum value name in CPP. For example: `UnSubAckReasonCode ::AWS_MQTT5_UARC_SUCCESS` -> `UnSubAckReasonCode::SUCCESS`. In this case, we could still directly pass `AWS_MQTT5_UARC_SUCCESS ` as the enum would be converted implicitly. However, it would break the user who was using `UnSubAckReasonCode::AWS_MQTT5_UARC_SUCCESS`. 

I'm not sure if there is a better way to fix it.

bretambrose: That was what I intended:

enum class UnsubackReasonCode {
   Success = AWS_MQTT5_UARC_SUCCESS,
   ... etc...
};

xiazhvera: 1. It seems that enum class would not be able to directly convert between int and enum value. Using enum class would definitely breaks the customer. 

2. If we removed AWS_MQTT5_UARC_SUCCESS and rename it to `Success`, then it would break the customer if they already use the enum with the namespace like following pattern:
```cpp
UnSubAckReasonCode reasonCode = UnSubAckReasonCode::AWS_MQTT5_UARC_SUCCESS
```

bretambrose: You could double the entries with the palatable names, but I'm not sure the extra effort is worth it though.

Forcing people to use the name mangling prefix is crummy  though.

include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -595,15 +595,27 @@ namespace Aws
                  */
                 Mqtt5ClientOptions &WithConnackTimeoutMs(uint32_t connackTimeoutMs) noexcept;
 
+                /**
+                 * @deprecated The function is deprecated, please use `Mqtt5ClientOptions::WithAckTimeoutSec(uint32_t)`
```
bretambrose, line 599: I'm confused.  Why are we adding a function and deprecating it at the same time?

source/iot/Mqtt5Client.cpp
```diff
@@ -408,9 +408,15 @@ namespace Aws
             return *this;
         }
 
-        Mqtt5ClientBuilder &Mqtt5ClientBuilder::WithAckTimeoutSeconds(uint32_t ackTimeoutSeconds) noexcept
+        Mqtt5ClientBuilder &Mqtt5ClientBuilder::WithAckTimeoutSec(uint32_t ackTimeoutSec) noexcept
         {
-            m_options->WithAckTimeoutSeconds(ackTimeoutSeconds);
+            m_options->WithAckTimeoutSec(ackTimeoutSec);
+            return *this;
+        }
```
bretambrose, line 415: Even though it's trivial, let's use a rule that the deprecated function should always just call the non-deprecated variant rather than repeat what the non-deprecated variant does.

include/aws/crt/mqtt/Mqtt5Client.h
bretambrose: Ah nm, diff was confusing

bretambrose (approved):

PR #584, opened 2023-12-13
Custom auth polish

bretambrose:
* Encode token signature if needed
* Update documentation


source/iot/Mqtt5Client.cpp
```diff
@@ -729,7 +730,17 @@ namespace Aws
 
         Mqtt5CustomAuthConfig &Aws::Iot::Mqtt5CustomAuthConfig::WithTokenSignature(Crt::String tokenSignature)
         {
-            m_tokenSignature = std::move(tokenSignature);
+            if (tokenSignature.find('%') != tokenSignature.npos)
+            {
+                // We can assume that a base 64 value that contains a '%' character has already been uri encoded
+                m_tokenSignature = tokenSignature;
```
sfod, line 736: nit: `std::move(tokenSignature)`

source/iot/Mqtt5Client.cpp
```diff
@@ -729,7 +730,17 @@ namespace Aws
...
+            }
+            else
+            {
+                m_tokenSignature =
+                    Aws::Crt::Io::EncodeQueryParameterValue(aws_byte_cursor_from_c_str(tokenSignature.c_str()));
```
sfod, line 741: Debatable nit: if change EncodeQueryParameterValue to accept `const String&`, the calling code will be less verbose

source/iot/MqttClient.cpp
```diff
@@ -381,8 +382,21 @@ namespace Aws
...
+                }
+                else
+                {
+                    encodedSignature = Aws::Crt::Io::EncodeQueryParameterValue(
+                        aws_byte_cursor_from_c_str(authorizerSignature.c_str()));
+                }
+
                 usernameString =
-                    AddToUsernameParameter(usernameString, authorizerSignature, "x-amz-customauthorizer-signature=");
+                    AddToUsernameParameter(usernameString, encodedSignature, "x-amz-customauthorizer-signature=");
```
sfod, line 399: nit: `std::move(usernameString)` and `std::move(encodedSignature)`

sfod (approved):

source/iot/MqttClient.cpp
```diff
@@ -381,8 +382,21 @@ namespace Aws
...
             }
             if (!tokenKeyName.empty() || !tokenValue.empty())
```
sfod, line 401: Similar to aws-crt-nodejs, we can check for a case when tokenKeyName and tokenValue are set while authorizerSignature is empty.

source/iot/Mqtt5Client.cpp
bretambrose: From a utility perspective, taking a byte cursor is quite a bit more useful.

PR #599, opened 2024-02-22
Basic version of Variant type using C++11

SergeyRyabinin:
*Issue #, if available:*
for C++ SDK
*Description of changes:*
Adding a basic version of a type/container Variant that represents a very basic version of std::variant but using C++11 and disabled exceptions.


include/aws/crt/Variant.h
```diff
@@ -0,0 +1,348 @@
...
+            using VariantAlternative = VariantAlternative<Index, Variant<Ts...>>;
+
+            using ThisVariantT = Variant<Ts...>;
+            template<typename OtherT>
+            using EnableIfOtherIsSameVariantType = typename std::enable_if<std::is_same<typename std::decay<OtherT>, ThisVariantT>::value, int>::type;
+
+            template<typename OtherT>
+            using EnableIfOtherIsThisVariantAlternative = typename std::enable_if<VariantDetail::Checker::HasType<typename std::decay<OtherT>::type, Ts...>::value, int>::type;
+
+            Variant() : m_value(nullptr) {}
```
graebm, line 111: docs say:
> Constructs a variant holding the [value-initialized](https://en.cppreference.com/w/cpp/language/value_initialization) value of the first alternative ([index()](https://en.cppreference.com/w/cpp/utility/variant/index) is zero).

so we need to set m_index=0, and need to actually run a constructor

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,348 @@
...
+            ~Variant()
+            {
+                Destroy();
+            }
+
+          private:
+            static constexpr std::size_t STORAGE_SIZE = VariantDetail::ParameterPackSize::GetMaxSizeOf<Ts...>();
+
+            alignas(Ts...) char m_storage[STORAGE_SIZE];
+            void *m_value = nullptr;
```
graebm, line 254: debatable: is it worth having `m_value`? In Optional, we needed SOME extra variable to track whether or not it holds a value, but Variant is required to be holding a value. We could have used a bool. My memory's not good enough to know whether we chose a pointer over a bool for alignment reasons, or just because it made our code/debugging simpler.

But you could get away without using it at all for Variant, and just using reinterpret_casts

tests/VariantTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        {
+            // jsut a different order or types
+            using MyTestVariant2 = Aws::Crt::Variant<Aws::Crt::String, int, char>;
+            MyTestVariant2 var2;
+            var2 = var2;
+            MyTestVariant2 var2CpyAssigned;
+            var2CpyAssigned = var2;
+            MyTestVariant2 var2CpyConstructedVariant(var2CpyAssigned);
```
graebm, line 40: Probably good to test move-constructors, and move-assignment too

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,348 @@
...
+                template<typename T, typename First, typename Second, typename ...Rest>
+                constexpr std::size_t GetIndexOf(std::size_t curIndex = 0) {
+                    return std::is_same<T, First>::value ? curIndex : GetIndexOf<T, Second, Rest...>(curIndex++);
+                }
+            }
+
+            namespace Checker
+            {
+                // Returns True if the template parameter pack Ts has a type T, i.e. ContainsType<T, Ts>() == true if T is in the list of Ts
+//                template<typename T, typename First, typename Second, typename ...Rest>
```
sbiscigl, line 55: nit: commented out code that is uncommented further down

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,348 @@
...
+            {
+                auto otherIdx = other.m_index;
+                if (otherIdx != -1)
+                {
+                    CopyAssignUtil(other);
+                }
+            }
+
+            template<typename VariantT = ThisVariantT, EnableIfOtherIsSameVariantType<VariantT> = 1>
+            Variant(VariantT &&other)
```
sbiscigl, line 125: should we consider using universal references? seems like a lot of this is split between a rvalue and a lvalue path. I too dont love universal refs but it feels like it could be used here

sbiscigl:
any plan on adding a a version of [std::visit](https://en.cppreference.com/w/cpp/utility/variant/visit) to this? I think if we added it alongside variant a whole new realm of static polymorphism opens up to us.

tests/VariantTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+            MyTestVariant1 var1CpyAssigned;
+            var1CpyAssigned = var1;
+            MyTestVariant1 var1CpyConstructedVariant(var1CpyAssigned);
+
+            MyTestVariant1 var1a = Aws::Crt::String(s_variant_test_str);
+            var1a = var1a;
+            var1 = var1a;
+            MyTestVariant1 var1aCpyAssigned;
+            var1CpyAssigned = var1a;
+            MyTestVariant1 var1aCpyConstructedVariant(var1aCpyAssigned);
```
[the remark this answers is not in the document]
SergeyRyabinin: Done, added more tests.

tests/VariantTest.cpp
SergeyRyabinin: Done, also added in-place c-tor option (and a test).

sbiscigl (approved):

tests/CMakeLists.txt
```diff
@@ -274,6 +274,8 @@ if(AWS_WARNINGS_ARE_ERRORS)
     endif()
 endif()
 
+target_compile_definitions(${TEST_BINARY_NAME} PRIVATE $<$<CONFIG:Debug>:DEBUG_BUILD>)
```
SergeyRyabinin, line 277: it was missing for the test project

sbiscigl (approved):

include/aws/crt/Variant.h
SergeyRyabinin: Done, removed m_value, 
still in places where the result of reinterpret cast is returned - a temp local variable is used to satisfy old gcc versions complaining about type strict aliasing.

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+
+            template <typename T, EnableIfOtherIsThisVariantAlternative<T> = 1> bool holds_alternative() const
+            {
+                AWS_ASSERT(m_index != -1);
+                return m_index == VariantDetail::Index::GetIndexOf<T, Ts...>();
+            }
+
+            template <size_t Index> constexpr bool holds_alternative() const { return Index < AlternativeCount; }
+
+            template <typename T> bool holds_alternative() const { return false; }
```
graebm, line 285: I am not a master of template programming. Should this be: `template <typename T, EnableIfOtherIsThisVariantAlternative<T> = 0> bool holds_alternative() const`???  I see this is the fallback to the other `holds_alternative()` that uses `enable_if`. Is this how you're supposed to implement a fallback? Or is it supposed to be more explicit?

But also, [docs](https://en.cppreference.com/w/cpp/utility/variant/holds_alternative) say:
> The call is ill-formed if T does not appear exactly once in Types...

so maybe we shouldn't even implement this?

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+
+            /* non-const get */
+            template <typename T, EnableIfOtherIsThisVariantAlternative<T> = 1> T &get()
+            {
+                T *value = reinterpret_cast<T *>(m_storage);
```
graebm, line 290: [docs](https://en.cppreference.com/w/cpp/utility/variant/get) say to throw `std::bad_variant_access` if T isn't what's currently stored here. But we compile without exceptions, right? In that case, what's the right thing to do? crash? If so, maybe `AWS_FATAL_ASSERT(holds_alternative<T>())`?

(AWS_FATAL_ASSERT is our exists-in-debug-and-release assert)

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+                static_assert(Index < AlternativeCount, "Unknown alternative index to emplace");
+                using AlternativeT = typename ThisVariantAlternative<Index>::type;
+
+                return emplace<AlternativeT, Args...>(args...);
+            }
+
+            template <typename T, EnableIfOtherIsThisVariantAlternative<T> = 1> bool holds_alternative() const
+            {
+                AWS_ASSERT(m_index != -1);
+                return m_index == VariantDetail::Index::GetIndexOf<T, Ts...>();
```
graebm, line 280: The [docs](https://en.cppreference.com/w/cpp/utility/variant/holds_alternative) say:
> The call is ill-formed if T does not appear exactly once in Types...

I genuinely don't know what "ill formed" means. Does that mean we should fail at compile time if T occurs twice in Types? Or does it mean that the implementation can do whatever it wants?

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+                }
+            }
+
+            template <typename T, EnableIfOtherIsThisVariantAlternative<T> = 1> Variant(const T &val)
+            {
+                static_assert(
+                    VariantDetail::Checker::HasType<typename std::decay<T>::type, Ts...>::value,
+                    "This variant does not have such alternative T.");
+                static_assert(
+                    sizeof(T) >= STORAGE_SIZE,
```
sfod, line 172: It seems, `<=` should be here

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+            Variant(Variant &&other)
+            {
+                auto otherIdx = other.m_index;
+                if (otherIdx != -1)
+                {
+                    MoveConstructUtil<0, Ts...>(std::move(other));
+                }
+            }
+
+            template <typename T, EnableIfOtherIsThisVariantAlternative<T> = 1> Variant(const T &val)
```
sfod, line 166: Debatable: This constructor is not needed, as the constructor with a forwarding reference parameter covers const references.

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+                return *value;
+            }
+
+            template <std::size_t Index, typename... Args>
+            auto emplace(Args &&...args) -> typename ThisVariantAlternative<Index>::type &
+            {
+                static_assert(Index < AlternativeCount, "Unknown alternative index to emplace");
+                using AlternativeT = typename ThisVariantAlternative<Index>::type;
+
+                return emplace<AlternativeT, Args...>(args...);
```
sfod, line 274: `args` should be forwarded here, i.e. `std::forward<Args>(args)...`.
Also, there are no tests for this version of `emplace`.

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,559 @@
...
+
+          private:
+            static constexpr std::size_t STORAGE_SIZE = VariantDetail::ParameterPackSize::GetMaxSizeOf<Ts...>();
+
+            alignas(VariantDetail::ParameterPackSize::AlignAsPack<Ts...>()) char m_storage[STORAGE_SIZE];
+            VariantDetail::Index::VariantIndex m_index = -1;
+
+            void Destroy()
+            {
+                if (m_index != -1)
```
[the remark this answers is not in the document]
SergeyRyabinin: I tried to eliminate all cases where user can set -1 type in the variant.
There are still corner cases (explicit d-tor call) or UB where this may happen - I've added asserts or fatal asserts accordingly.

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+            ~Variant() { Destroy(); }
+
+          private:
+            static constexpr std::size_t STORAGE_SIZE = VariantDetail::ParameterPackSize::GetMaxSizeOf<Ts...>();
+
+            alignas(VariantDetail::ParameterPackSize::AlignAsPack<Ts...>()) char m_storage[STORAGE_SIZE];
+            VariantDetail::Index::VariantIndex m_index = -1;
+#if defined(DEBUG_BUILD)
+            VariantDetail::VariantDebug::VariantDebugBrowser<Ts...> browser = m_storage;
+#endif /* defined(DEBUG_BUILD) */
```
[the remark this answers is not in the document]
SergeyRyabinin: I've changed the macro variable to be more specific and dedicated `AWS_CRT_ENABLE_VARIANT_DEBUG`, it needs to be explicitly enabled.

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,566 @@
...
+            {
+                static_assert(Index < AlternativeCount, "Attempting to construct unknown Index Type");
+
+                if (Index == other.m_index)
+                {
+                    using AlternativeT = typename ThisVariantAlternative<Index>::type;
+                    new (m_storage) AlternativeT(other.get<Index>());
+                    m_index = Index;
+                    AWS_ASSERT(m_index != -1);
+                }
```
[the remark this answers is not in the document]
SergeyRyabinin: I've implemented the visitor and extracted repeated code.
There might be some more for further polishing, but my brain hurts even at this point.

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+            template <std::size_t Index> using ThisVariantAlternative = VariantAlternative<Index, Ts...>;
+
+            using ThisVariantT = Variant<Ts...>;
+            template <typename OtherT>
+            using EnableIfOtherIsSameVariantType =
+                typename std::enable_if<std::is_same<typename std::decay<OtherT>, ThisVariantT>::value, int>::type;
+
+            template <typename OtherT>
+            using EnableIfOtherIsThisVariantAlternative = typename std::
+                enable_if<VariantDetail::Checker::HasType<typename std::decay<OtherT>::type, Ts...>::value, int>::type;
```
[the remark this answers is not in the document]
SergeyRyabinin: moved to private, thanks.

include/aws/crt/Variant.h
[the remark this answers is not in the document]
SergeyRyabinin: As mentioned above, I've switched to a completely different macro variable name, not connected with debug build and to be enabled explicitly if needed.

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+                return emplace<AlternativeT, Args...>(args...);
+            }
+
+            template <typename T, EnableIfOtherIsThisVariantAlternative<T> = 1> bool holds_alternative() const
+            {
+                AWS_ASSERT(m_index != -1);
+                return m_index == VariantDetail::Index::GetIndexOf<T, Ts...>();
+            }
+
+            template <size_t Index> constexpr bool holds_alternative() const { return Index < AlternativeCount; }
```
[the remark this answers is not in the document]
SergeyRyabinin: good catch, thank you

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+                new (m_storage) FirstAlternative();
+                m_index = 0;
+            }
+
+            Variant(const Variant &other)
+            {
+                auto otherIdx = other.m_index;
+                if (otherIdx != -1)
+                {
+                    CopyConstructUtil<0, Ts...>(other);
```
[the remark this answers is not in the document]
SergeyRyabinin: as mentioned above, m_index is always initialized to -1 first.
I've added more asserts and fatal asserts accordingly.
I don't know if there are more places to add error handling asserts, the time will tell.

include/aws/crt/Variant.h
SergeyRyabinin: A good catch!
I've moved it to private cause it is pretty handy.
Also, I've removed generic version for types not present in Variant (as it does not exist for std variant).

include/aws/crt/Variant.h
SergeyRyabinin: ill formed means, [thanks to this stackoverflow reply](https://stackoverflow.com/a/15805532)
> If the Standard does not specify otherwise, an implementation should emit a diagnostic message (error or warning).

It is an undefined behavior. It is not very well specified how to handle repeated types in a variant. This variant implementation always visits left-to-right (except when directly accessed by Index, it is an UB in such case).

include/aws/crt/Variant.h
SergeyRyabinin: Good catch, thank you!
However, my brain hurts more trying to think more about this one. I'd prefer to keep it as is as of now.

include/aws/crt/Variant.h
SergeyRyabinin: Good catch, thank you

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,595 @@
...
+                    VariantDetail::Checker::HasType<typename std::decay<T>::type, Ts...>::value,
+                    "This variant does not have such alternative T.");
+                static_assert(
+                    sizeof(T) <= STORAGE_SIZE,
+                    "Attempting to instantiate a Variant with a type bigger than all alternatives.");
+
+                using PlainT = typename std::decay<T>::type;
+                new (m_storage) PlainT(std::forward<T>(val));
+                m_index = VariantDetail::Index::GetIndexOf<T, Ts...>();
+                AWS_ASSERT(m_index != -1);
```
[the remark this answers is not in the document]
SergeyRyabinin: Good catch, thank you

sbiscigl (approved):

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,588 @@
...
+
+                template <typename AlternativeT, typename ConstAlternativeT>
+                void operator()(AlternativeT &&value, ConstAlternativeT &other) const
+                {
+                    using PlaintT = typename std::remove_reference<AlternativeT>::type;
+                    using PlaintOtherT =
+                        typename std::remove_const<typename std::remove_reference<AlternativeT>::type>::type;
+                    static_assert(std::is_same<PlaintT, PlaintOtherT>::value, "Incompatible types");
+
+                    value = std::move(other);
```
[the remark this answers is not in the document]
SergeyRyabinin: right, std::move just did not work on a const type, so "technically no issue".... therefore it is not caught by a test.

Thanks a lot for catching it!

include/aws/crt/Variant.h
```diff
@@ -0,0 +1,588 @@
...
+                        VisitorUtil<0, Ts...>::VisitBinary(this, other, CopyMoveAssigner());
+                    }
+                }
+                return *this;
+            }
+
+            Variant &operator=(Variant &&other)
+            {
+                AWS_ASSERT(other.m_index != -1);
+                if (this != &other && other.m_index != -1)
```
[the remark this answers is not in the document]
SergeyRyabinin: well, okay, I agree with such change, the only change I made was to
```
                if (this != &other)
                {
                    AWS_FATAL_ASSERT(other.m_index != -1);
```
in operators=
(constructors will fail unconditionally)

graebm (approved):

PR #600, opened 2024-02-26
Add in-place constructor and emplace method to Optional

SergeyRyabinin:
*Issue #, if available:*
NOW there is no way to avoid copy or move construction of underlying value
*Description of changes:*
Provide such option in Optional.


sbiscigl (approved):

tests/OptionalMemorySafetyTest.cpp
```diff
@@ -56,6 +56,54 @@ static int s_OptionalMoveSafety(struct aws_allocator *allocator, void *ctx)
...
+size_t EmplaceTester::ctorCallCount = 0;
+size_t EmplaceTester::dtorCallCount = 0;
+
+static int s_OptionalEmplace(struct aws_allocator *allocator, void *ctx)
+{
+    (void)ctx;
+    {
+        Aws::Crt::ApiHandle apiHandle(allocator);
+
+        Aws::Crt::Optional<Aws::Crt::String> str1{Aws::Crt::Optional<Aws::Crt::String>::in_place, s_test_str};
```
graebm, line 83: [debatable] Maybe worth declaring `in_place` outside the class (like it is in std)? That would make using it shorter to type. We could add an `include/aws/crt/Utility.h` file so both Optional.h and Variant.h could use it:
```suggestion
        Aws::Crt::Optional<Aws::Crt::String> str1{Aws::Crt::InPlace, s_test_str};
```

Or we could keep it in `Optional.h`, but keep it outside the `Optional<T>` class, like:
```suggestion
        Aws::Crt::Optional<Aws::Crt::String> str1{Aws::Crt::OptionalInPlace, s_test_str};
```

SergeyRyabinin: Thank you for a suggested change, I've updated the PR to do this.
(I'll rebase Variant PR after merging this PR to use the Utility header)

graebm (approved):

PR #617, opened 2024-04-30
Fix memory leak when UnsubscribePacket is reused

sfod:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/mqtt/Mqtt5Packets.cpp
+++ b/source/mqtt/Mqtt5Packets.cpp
@@ -138,7 +138,7 @@ namespace Aws
                 const Crt::Vector<String> &stringVector,
                 Allocator *allocator)
             {
-                AWS_ZERO_STRUCT(dst);
+                aws_array_list_clean_up(&dst);
 
                 if (aws_array_list_init_dynamic(&dst, allocator, stringVector.size(), sizeof(aws_byte_cursor)) !=
                     AWS_OP_SUCCESS)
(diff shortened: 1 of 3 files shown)


xiazhvera (approved):

bretambrose (commented):
I feel like there may be larger issues.  What happens when an  Unsubscribe gets submitted by multiple threads.  (Unsafely) Caching ephemeral data on the packet right before we submit to the native client feels broken.

sfod:
> Caching ephemeral data on the packet right before we submit to the native client feels broken.

Unfortunately, the existing API of the `UnsubscribePacket` doesn't give much options on how to fix this issue:

```c++
bool initializeRawOptions(aws_mqtt5_packet_unsubscribe_view &raw_options) noexcept;
```

I wanted to introduce new class and method, something like this:

```c++

// Cleans up unsubscribeView on destruction.
class AwsMqtt5PacketUnsubscribeView_WrapperRAII {
    // ctor and dtor with RAII

    aws_mqtt5_packet_unsubscribe_view unsubscribeView;
};

AwsMqtt5PacketUnsubscribeView_WrapperRAII UnsubscribePacket::getRawOptions() noexcept;
```

, then the calling code would look like this:

```c++
bool Mqtt5ClientCore::Unsubscribe()
{
    ... 
    auto unsubscribeViewWrapper = unsubscribePacket->getRawOptions();
    int result = aws_mqtt5_client_unsubscribe(m_client, &unsubscribeViewWrapper.unsubscribeView, &options);
    return result;
}
```

But it'll be backward-incompatible change for `UnsubscribePacket`.

bretambrose:
Why not ditch (ignore initializeFromRawOptions) and use something new instead?

sfod:
`initializeFromRawOptions` has to remain valid, and it depends on the "cached" data inside the `UnsubscribePacket`, so we can't remove these fields, right?
Having two independent methods performing the same logic by two different ways feels wrong.

PR #627, opened 2024-06-12
cbor binding

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- Bare minimal binding for cbor
- follow up:
  - Overload to support std::string for text and bytes?
  - Array and map support from stdlib?
  - Time stamp support? `std::chrono::system_clock::time_point`?

- We will need those follow up for sure, but debatable about where they should live, maybe in C++ SDK? Or maybe in crt-ccp, we should hold for the follow-ups until C++ SDK picks up the integration job

~Won't merge until the C code merge to main and the follow up has been addressed~


include/aws/crt/cbor/Cbor.h
```diff
@@ -0,0 +1,362 @@
...
+                Null = AWS_CBOR_TYPE_NULL,
+                Undefined = AWS_CBOR_TYPE_UNDEFINED,
+                Break = AWS_CBOR_TYPE_BREAK,
+                IndefBytesStart = AWS_CBOR_TYPE_INDEF_BYTES_START,
+                IndefTextStart = AWS_CBOR_TYPE_INDEF_TEXT_START,
+                IndefArrayStart = AWS_CBOR_TYPE_INDEF_ARRAY_START,
+                IndefMapStart = AWS_CBOR_TYPE_INDEF_MAP_START,
+            };
+
+            class AWS_CRT_CPP_API CborEncoder
```
SergeyRyabinin, line 56: add final if `destructor` is not `virtual`

SergeyRyabinin (approved):

source/cbor/Cbor.cpp
```diff
@@ -0,0 +1,260 @@
...
+        namespace Cbor
+        {
+            /*****************************************************
+             *
+             * CborEncoder
+             *
+             *****************************************************/
+            CborEncoder::CborEncoder(Crt::Allocator *allocator) noexcept
+            {
+                m_encoder = aws_cbor_encoder_new(allocator);
```
graebm, line 21: there's no possible way that aws_cbor_encoder_new() or aws_cbor_decoder_new() could ever return NULL in the future, right? right?

if so, we won't want to use normal C++ constructors, because they can't fail without the use of exceptions, which we turn off

include/aws/crt/cbor/Cbor.h
```diff
@@ -0,0 +1,362 @@
...
+
+                /**
+                 * Consume the next data item, includes all the content within the data item.
+                 *
+                 * As an example for the following CBOR, this function will consume all the data
+                 * as it's only one CBOR data item, an indefinite map with 2 key, value pair:
+                 * 0xbf6346756ef563416d7421ff
+                 * BF           -- Start indefinite-length map
+                 *   63        -- First key, UTF-8 string length 3
+                 *      46756e --   "Fun"
```
graebm, line 254: utterly trivial. ascii art alignment
```suggestion
                 * BF          -- Start indefinite-length map
                 *   63        -- First key, UTF-8 string length 3
                 *      46756e --   "Fun"
```

include/aws/crt/cbor/Cbor.h
```diff
@@ -0,0 +1,362 @@
...
+                /**
+                 * Consume the next single element, without the content followed by the element.
+                 *
+                 * As an example for the following CBOR, this function will only consume the
+                 * 0xBF, "Start indefinite-length map", not any content of the map represented.
+                 * The next element to decode will start from 0x63.
+                 * 0xbf6346756ef563416d7421ff
+                 * BF           -- Start indefinite-length map
+                 *   63        -- First key, UTF-8 string length 3
+                 *      46756e --   "Fun"
```
graebm, line 276:
```suggestion
                 * BF          -- Start indefinite-length map
                 *   63        -- First key, UTF-8 string length 3
                 *      46756e --   "Fun"
```

include/aws/crt/cbor/Cbor.h
```diff
@@ -0,0 +1,362 @@
...
+                size_t GetRemainingLength() noexcept;
+
+                /**
+                 * Decode the next element and store it in the decoder cache if there was no element cached.
+                 * If there was an element cached, just return the type of the cached element.
+                 *
+                 * @param out_type
+                 * @return success/failure
+                 */
+                bool PeekType(CborType &out_type) noexcept;
```
graebm, line 244: For all the functions with out-types, there are nicer ways to do stuff like this in C++
```suggestion
                Optional<CborType> PeekType() noexcept;
```

include/aws/crt/cbor/Cbor.h
```diff
@@ -0,0 +1,362 @@
...
+                 * @return The length of bytes remaining of the decoder source.
+                 */
+                size_t GetRemainingLength() noexcept;
+
+                /**
+                 * Decode the next element and store it in the decoder cache if there was no element cached.
+                 * If there was an element cached, just return the type of the cached element.
+                 *
+                 * @param out_type
+                 * @return success/failure
```
graebm, line 242: document what the user should do in the case of failure. We never really settled on a standardized way to do last-error in aws-crt-cpp, so you need to tell people how THIS class handles it

source/cbor/Cbor.cpp
TingDaoK: I don't think we will error let the new function fail for the encoder and decoder, besides of allocation failure

include/aws/crt/cbor/Cbor.h
```diff
@@ -0,0 +1,367 @@
...
+                 * @return If successful, return the tag value
+                 *         If not, return will be none and Aws::Crt::LastError() can be
+                 *          used to retrieve CRT error code.
+                 */
+                Optional<uint64_t> PopNextTagVal() noexcept;
+
+                /**
+                 * @return the value of the last aws error encountered by operations on this instance.
+                 */
+                int LastError() const noexcept { return m_lastError ? m_lastError : AWS_ERROR_UNKNOWN; }
```
graebm, line 357: You have this function, but wrote all the other docs to say "If not, return will be none and Aws::Crt::LastError() can be used to retrieve CRT error code."

Alter docs to tell users that they should use it, or just delete this (and m_lastError)

graebm (approved):
fix & ship

PR #629, opened 2024-06-16
Fix ASan report about alloc/dealloc mismatch when compiled with C++23

alexey-milovidov:
*Description of changes:*

When the library is compiled with -std=c++23, ASan starts to complain about mismatched allocation/deallocation, like the following:

```
2024-06-16 08:26:29 =================================================================
2024-06-16 08:26:29 ==2168==ERROR: AddressSanitizer: alloc-dealloc-mismatch (operator new vs free) on 0x506000109040
2024-06-16 08:26:29     #0 0x558a1c76a236 in free (/usr/bin/clickhouse+0xa4fb236) (BuildId: 9a4ef85e15c9099607afed77eac4c9d0390ac9f2)
2024-06-16 08:26:29     #1 0x558a45097fd4 in std::__1::vector<aws_host_address, Aws::Crt::StlAllocator<aws_host_address>>::push_back[abi:v15000](aws_host_address const&) HostResolver.cpp
2024-06-16 08:26:29     #2 0x558a45097a14 in Aws::Crt::Io::DefaultHostResolver::s_onHostResolved(aws_host_resolver*, aws_string const*, int, aws_array_list const*, void*) (/usr/bin/clickhouse+0x32e28a14) (BuildId: 9a4ef85e15c9099607afed77eac4c9d0390ac9f2)
2024-06-16 08:26:29     #3 0x558a45092762 in aws_host_resolver_thread host_resolver.c
2024-06-16 08:26:29     #4 0x558a44fe282c in thread_fn thread.c
2024-06-16 08:26:29     #5 0x558a1c768058 in asan_thread_start(void*) crtstuff.c
2024-06-16 08:26:29     #6 0x7f719ff95ac2  (/lib/x86_64-linux-gnu/libc.so.6+0x94ac2) (BuildId: 962015aa9d133c6cbcfb31ec300596d7f44d3348)
2024-06-16 08:26:29     #7 0x7f71a002784f  (/lib/x86_64-linux-gnu/libc.so.6+0x12684f) (BuildId: 962015aa9d133c6cbcfb31ec300596d7f44d3348)
2024-06-16 08:26:29 
2024-06-16 08:26:29 0x506000109040 is located 0 bytes inside of 64-byte region [0x506000109040,0x506000109080)
2024-06-16 08:26:29 allocated by thread T3 here:
[...]

PR #633, opened 2024-06-24
Fix ASan report about alloc/dealloc mismatch when compiled with C++23

waahm7:
*Issue #, if available:*

*Description of changes:*
Duplicate of #629 to fix lint and submodules


include/aws/crt/StlAllocator.h
graebm, line 1: If `std::allocator<T>` doesn't have any virtual functions ... does that mean nothing should inherit from it?

waahm7: Discussed offline, we will backlog this for now.

include/aws/crt/StlAllocator.h
```diff
@@ -51,6 +51,10 @@ namespace Aws
                 return static_cast<RawPointer>(aws_mem_acquire(m_allocator, n * sizeof(T)));
             }
 
+#if _LIBCPP_STD_VER > 20
```
DmitriyMusatkin, line 54: do we really need #if here? 
does it really matter if we expose it in all versions of cpp standard?

waahm7: std::allocation_result was added in _LIBCPP_STD_VER > 20

DmitriyMusatkin (approved):

PR #644, opened 2024-07-31
Fix HTTP proxy bug in Mqtt5ClientBuilder

elryno:
The `Mqtt5ClientBuilder` does not properly set the `Mqtt5ClientOptions` `m_options` field in the non-WebSocket case of the `Build()` function. Handle this case.


bretambrose:
Merged in https://github.com/awslabs/aws-crt-cpp/pull/647

PR #664, opened 2024-09-19
Request response workspace

bretambrose:
Bindings for MQTT request-response client and associated types


include/aws/iot/MqttRequestResponseClient.h
```diff
@@ -0,0 +1,378 @@
...
+              public:
+                IncomingPublishEvent() : m_payload() { AWS_ZERO_STRUCT(m_payload); }
+                IncomingPublishEvent(const IncomingPublishEvent &rhs) = default;
+                IncomingPublishEvent(IncomingPublishEvent &&rhs) = default;
+                ~IncomingPublishEvent() = default;
+
+                IncomingPublishEvent &operator=(const IncomingPublishEvent &rhs) = default;
+                IncomingPublishEvent &operator=(IncomingPublishEvent &&rhs) = default;
+
+                IncomingPublishEvent &WithPayload(Aws::Crt::ByteCursor payload)
```
xiazhvera, line 111: For mqtt5 client, we keep a copy of the buffer data. Not sure which way is a better approach though.

bretambrose: Why make a copy unless it's necessary?

include/aws/iot/MqttRequestResponseClient.h
```diff
@@ -0,0 +1,378 @@
...
+              public:
+                SubscriptionStatusEvent() : m_type(SubscriptionStatusEventType::SubscriptionEstablished), m_errorCode(0)
+                {
+                }
+                SubscriptionStatusEvent(const SubscriptionStatusEvent &rhs) = default;
+                SubscriptionStatusEvent(SubscriptionStatusEvent &&rhs) = default;
+                ~SubscriptionStatusEvent() = default;
+
+                SubscriptionStatusEvent &operator=(const SubscriptionStatusEvent &rhs) = default;
+                SubscriptionStatusEvent &operator=(SubscriptionStatusEvent &&rhs) = default;
```
sfod, line 75: nit: These 5 special members will be generated by the compiler. A user-defined default constructor doesn't prevent generating them.

include/aws/iot/MqttRequestResponseClient.h
```diff
@@ -0,0 +1,378 @@
...
+                }
+
+                Result &operator=(R &&response)
+                {
+                    this->m_rawResult = std::move(response);
+
+                    return *this;
+                }
+
+                Result &operator=(const E &error) { this->m_rawResult = error; }
```
sfod, line 205: Missing `return *this;`

source/iot/MqttRequestResponseClient.cpp
```diff
@@ -0,0 +1,509 @@
...
+            {
+              public:
+                StreamReadLock(struct aws_rw_lock *lock, struct aws_event_loop *protocolLoop)
+                    : m_lock(lock), m_taken(false)
+                {
+                    if (!aws_event_loop_thread_is_callers_thread(protocolLoop))
+                    {
+                        m_taken = true;
+                        aws_rw_lock_rlock(lock);
+                    }
```
sfod, line 31: I think this `if` logic might be flawed.
If `StreamReadLock` is created by the event loop's thread, it won't lock `lock`. So, some writer from another thread will be able to obtain `lock` and get access to protected data while event loop thread is reading it.

source/iot/MqttRequestResponseClient.cpp
```diff
@@ -0,0 +1,509 @@
...
+                Aws::Crt::Allocator *m_allocator;
+
+                std::shared_ptr<StreamingOperationImpl> m_impl;
+            };
+
+            StreamingOperationImpl::StreamingOperationImpl(
+                struct aws_mqtt_rr_client_operation *stream,
+                const StreamingOperationOptionsInternal &options,
+                struct aws_event_loop *protocolLoop)
+                : m_config(options), m_stream(stream), m_protocolLoop(protocolLoop), m_lock(), m_closed(false)
```
sfod, line 122: nit: Maybe validate that `options` has all fields set? Because down the code we execute `m_config` members without checking them.

source/iot/MqttRequestResponseClient.cpp
bretambrose: Aw crud

source/iot/MqttRequestResponseClient.cpp
```diff
@@ -0,0 +1,509 @@
...
+            }
+
+            std::shared_ptr<IStreamingOperation> StreamingOperation::Create(
+                Aws::Crt::Allocator *allocator,
+                const StreamingOperationOptionsInternal &options,
+                struct aws_mqtt_request_response_client *client)
+            {
+                auto *implHandle = Aws::Crt::New<StreamingOperationImplHandle>(allocator);
+
+                struct aws_mqtt_streaming_operation_options streamingOptions;
```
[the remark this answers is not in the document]
bretambrose: I like to keep it on C types.

source/iot/MqttRequestResponseClient.cpp
bretambrose: Added function containment checks before the callbacks

source/iot/MqttRequestResponseClient.cpp
```diff
@@ -0,0 +1,493 @@
...
+                }
+                else
+                {
+                    s_completeRequestWithSuccess(incompleteRequest, response_topic, payload);
+                }
+
+                Aws::Crt::Delete(incompleteRequest, incompleteRequest->m_allocator);
+            }
+
+            class MqttRequestResponseClientImpl
```
xiazhvera, line 316: I'm not super clear what is purpose of `MqttRequestResponseClientImpl`. So far the `Impl` worked as a "MqttRequestResponseClientCore" which handles the releasing of the client. Meanwhile, from the name, I got the idea that we probably wanna extend the `MqttRequestResponseClient` with different client implementation. Would we have plan for different client implementation?

xiazhvera (commented):
I'm half way through the code review, and gona continue tomorrow. I have couple questions regards to the implementation and so I post first...

source/iot/MqttRequestResponseClient.cpp
```diff
@@ -0,0 +1,493 @@
...
+                struct aws_rw_lock *m_lock;
+            };
+
+            class StreamWriteLock
+            {
+              public:
+                StreamWriteLock(struct aws_rw_lock *lock, struct aws_event_loop *protocolLoop)
+                    : m_lock(lock), m_taken(false)
+                {
+                    if (!aws_event_loop_thread_is_callers_thread(protocolLoop))
```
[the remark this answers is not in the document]
sfod: Not related to Vera's comment, but again a concern about race condition between `StreamWriteLock` and `StreamReadLock`.

If `StreamReadLock` can be used outside the event loop thread, this `if` logic can cause the data race if `StreamWriteLock` is used in the event loop. Even if it's not possible at the moment, we should consider if this situations might appear in the future.

I think, if there is no evidence this optimization actually makes a noticeable difference, we should remove it.

include/aws/crt/mqtt/Mqtt5Client.h
```diff
@@ -435,6 +435,8 @@ namespace Aws
 
                 virtual ~Mqtt5Client();
 
+                struct aws_mqtt5_client *GetUnderlyingHandle() const noexcept;
```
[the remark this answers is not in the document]
bretambrose: My blunt response to that would be something like this: "that scenario represents such a perversely foolish/self-destructive behavior by the user that I have no interest in trying to prevent it."

More generally, getters for the underlying C object is a pattern we use throughout the CRT.

source/iot/MqttRequestResponseClient.cpp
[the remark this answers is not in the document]
bretambrose: StreamReadLock is only used when invoking the event-loop sourced callbacks.
It's also not an optimization; it's 100% required for correctness.  Without it you get a deadlock if the C++ request-response client is dropped by the callback.

Basic take away is that I need to document the purpose and usage of these types better.

source/iot/MqttRequestResponseClient.cpp
bretambrose: There are no plans for different implementations.  I just prefer "Impl" to "Core"

source/iot/MqttRequestResponseClient.cpp
```diff
@@ -0,0 +1,493 @@
...
+            void MqttRequestResponseClientImpl::Close() noexcept
+            {
+                aws_mqtt_request_response_client_release(m_client);
+                m_client = nullptr;
+            }
+
+            int MqttRequestResponseClientImpl::SubmitRequest(
+                const aws_mqtt_request_operation_options &requestOptions,
+                UnmodeledResultHandler &&resultHandler) noexcept
+            {
```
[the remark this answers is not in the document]
bretambrose: In what scenario could it be null?

include/aws/iot/MqttRequestResponseClient.h
```diff
@@ -0,0 +1,550 @@
...
+                /**
+                 * Duration, in seconds, that a request-response operation will wait for completion before giving up
+                 */
+                uint32_t m_operationTimeoutInSeconds = 0;
+            };
+
+            /**
+             * Generic interface for the request-response client
+             */
+            class AWS_CRT_CPP_API IMqttRequestResponseClient
```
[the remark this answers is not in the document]
bretambrose: It's internal only.  I'm not convinced that hiding it documentation-wise does any good.

tests/MqttRequestResponse.cpp
```diff
@@ -0,0 +1,1099 @@
...
+    Aws::Iot::RequestResponse::RequestResponseClientOptions *options = NULL)
+{
+    TestContext context;
+
+    Aws::Iot::RequestResponse::RequestResponseClientOptions finalOptions;
+    struct aws_string *host = NULL;
+    struct aws_string *certificatePath = NULL;
+    struct aws_string *privateKeyPath = NULL;
+
+    if (aws_get_environment_value(allocator, s_rrEnvVariableHost, &host))
```
sfod, line 230: `aws_get_environment_value` returns success on missing env var (kinda counterintuitive, I'd say).
As a side note: in the iot tests, I implemented a wrapper function returning `Optional<String>`: https://github.com/awslabs/aws-crt-cpp/blob/main/tests/IotServiceTest.cpp#L50

sfod (approved):

source/iot/MqttRequestResponseClient.cpp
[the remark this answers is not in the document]
bretambrose: I re-checked the usage.  It's usage is a little bit more general and also includes Open().  I don't see a race condition possible from there though.

source/iot/MqttRequestResponseClient.cpp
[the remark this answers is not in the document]
xiazhvera: In the scenario where someone called `Close` before call `SubmitRequest`. Regardless, it's not necessary, `aws_mqtt_request_response_client_submit_request` did a null check on m_client.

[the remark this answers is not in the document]
bretambrose: This isn't a public type.  I don't think that's possible.  Close is called by the public type's destructor.

include/aws/iot/MqttRequestResponseClient.h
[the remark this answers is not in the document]
xiazhvera: The API doc is public, so I don’t see a reason for exposing an internal class to the public. Anyway, it's trivial.

PR #686, opened 2024-11-27
Fix crash when building with _GLIBCXX_USE_CXX11_ABI=0

graebm:
**Issue:**
A user had been building aws-crt-cpp with `CXXFLAGS=-D_GLIBCXX_USE_CXX11_ABI=0`, and that worked fine until the recent change where we started hiding symbols by default (PR https://github.com/awslabs/aws-crt-cpp/pull/666)

With symbols hidden by default, and `_GLIBCXX_USE_CXX11_ABI=0`, tests would crash in the destructor of `Aws::Crt::String` (which is `std::string` with a custom allocator).

**Description of changes:**
Don't hide symbols when building for the ancient glibcxx ABI.

I don't 100% understand why this fixes the issue, but the situation is esoteric enough that it doesn't seem worth spending more days on this problem.

**Background**
See: https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html
GCC had to introduce a new ABI for C++11 which removed the copy-on-write optimization for strings, which is forbidden in C++11 and later (new ABI uses a small-string optimization instead). But GCC let users manually set `_GLIBCXX_USE_CXX11_ABI=0` so they could opt back into the old ABI and continue working with libraries compiled for C++03.

I don't think GCC intended devs to continue using this 10 years later, but some people are, because the old copy-on-write optimization has less memory usage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,9 +144,8 @@ jobs:
         ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ env.LINUX_BASE_IMAGE }} build -p ${{ env.PACKAGE_NAME }} --compiler=${{ matrix.compiler }} --cmake-extra=-DBUILD_SHARED_LIBS=ON
 
   linux-glibcxx-ancient-abi:
-    runs-on: ubuntu-22.04 # latest
+    runs-on: ubuntu-24.04 # latest
     steps:
-        # We can't use the `uses: docker://image` version yet, GitHub lacks authentication for actions -> packages
     - name: Build ${{ env.PACKAGE_NAME }}
       run: |
         aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ env:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-  
+
 jobs:
   linux-compat-use-openssl:
     runs-on: ubuntu-22.04 # latest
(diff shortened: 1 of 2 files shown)


TingDaoK (approved):

.github/workflows/ci.yml
```diff
@@ -143,6 +143,15 @@ jobs:
         aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
         ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ env.LINUX_BASE_IMAGE }} build -p ${{ env.PACKAGE_NAME }} --compiler=${{ matrix.compiler }} --cmake-extra=-DBUILD_SHARED_LIBS=ON
 
+  linux-glibcxx-ancient-abi:
+    runs-on: ubuntu-22.04 # latest
```
waahm7, line 147:
```suggestion
    runs-on: ubuntu-24.04 # latest
```

waahm7 (approved):

.github/workflows/ci.yml
```diff
@@ -143,6 +143,15 @@ jobs:
...
+    steps:
+        # We can't use the `uses: docker://image` version yet, GitHub lacks authentication for actions -> packages
```
waahm7, line 149: I feel like this comment has been copied and pasted so many times that it's not valuable anymore.
```suggestion
```

CMakeLists.txt
```diff
@@ -298,9 +298,17 @@ set_target_properties(${PROJECT_NAME} PROPERTIES LINKER_LANGUAGE CXX)
 set_target_properties(${PROJECT_NAME} PROPERTIES CXX_STANDARD ${CMAKE_CXX_STANDARD})
 
 # Hide symbols by default
-# Except for ancient GCC, because it leads to crashes in shared-lib builds
-# see: https://github.com/awslabs/aws-crt-cpp/pull/675
-if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "5.0"))
+# Except where it causes problems, and the situation is weird enough that it's not worth investigating further.
+string(FIND "${CMAKE_CXX_FLAGS}" "-D_GLIBCXX_USE_CXX11_ABI=0" found_ancient_abi_flag)
```
sfod, line 302: It's possible to set CXX flags via [`CMAKE_CXX_FLAGS_<CONFIG>`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG.html).

The following command will add `-D_GLIBCXX_USE_CXX11_ABI=0` option to compile command, but `CMAKE_CXX_FLAGS` will be empty:
```
cmake -S . -B build -DCMAKE_CXX_FLAGS_DEBUG="-D_GLIBCXX_USE_CXX11_ABI=0" -DCMAKE_BUILD_TYPE=Debug
```

sfod: I don't think developers actually set CXX flags this way, but additional comment might help to debug in the future if someone decides to do it.

graebm: Nice callout. Checking every variant `_<CONFIG>` variant now

I tried to be clever and use [CheckCXXSourceCompiles](https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html), but this didn't take the `_<CONFIG>` variants into account

sfod (approved):

PR #721, opened 2025-03-27
Capture fields from PUBLISH for req-resp stream operations

sfod:
Certain service clients need additional data from MQTT PUBLISH packet.
Currently, the following fields should be added:
- content type
- user properties
- message expiry interval seconds

In future, more data may be added.

Description of changes:
Pass all these fields to callback as a struct.


sbSteveK (approved):
update aws-c-mqtt to point to right release before merge

PR #722, opened 2025-03-27
Initialize checksums explicitly

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Call checksums module init explicitly to avoid any thread safety warnings.

USE_WINDOWS_DLL_SEMANTICS is renamed to AWS_CRT_USE_WINDOWS_DLL_SEMANTICS to be consistent with other repos
and removing some logic for gcc < 4, since we stopped supporting it long ago


SergeyRyabinin (approved):

source/Api.cpp
```diff
@@ -48,6 +49,7 @@ namespace Aws
             aws_s3_library_init(allocator);
             aws_event_stream_library_init(allocator);
             aws_sdkutils_library_init(allocator);
+            aws_checksums_library_init(allocator);
```
waahm7, line 52: technically we don't need to do this since aws-c-s3 will do it.

waahm7 (approved):

source/Api.cpp
DmitriyMusatkin: yes, but i think its better to be explicit here and just init what you use without relying on other things doing it for you.

PR #723, opened 2025-03-27
Fix IP address being labelled "bad" for too long

graebm:
Update submodules, bringing in this notable fix:
- https://github.com/awslabs/aws-c-io/pull/718

```
aws-c-auth         v0.8.6 -> v0.9.0
aws-c-cal          v0.8.7 -> v0.8.8
aws-c-common       v0.12.0 -> v0.12.2
aws-c-io           v0.17.0 -> v0.17.1
aws-checksums      v0.2.3 -> v0.2.5
aws-lc             v1.48.4 -> v1.48.5
s2n                v1.5.14 -> v1.5.15
```

### DIFF
--- a/source/Api.cpp
+++ b/source/Api.cpp
@@ -9,7 +9,6 @@
 #include <aws/crt/io/TlsOptions.h>
 
 #include <aws/auth/auth.h>
-#include <aws/checksums/checksums.h>
 #include <aws/common/ref_count.h>
 #include <aws/event-stream/event_stream.h>
 #include <aws/http/http.h>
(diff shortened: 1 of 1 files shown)


source/Api.cpp
```diff
@@ -44,6 +45,7 @@ namespace Aws
         {
             // sets up the StlAllocator for use.
             g_allocator = allocator;
+            aws_checksums_library_init(allocator);
```
TingDaoK, line 48: s3 depends on checksum, we should not need this?

graebm: 🤷‍♀️it's harmless to multiple-init, since the tree of dependencies isn't 100% clear, and aws-crt-cpp **does** directly call into checksums

but really, I need to get a release out today, and latest-submodules brings this fix, and I don't want to wait while I update several other libraries, and thought if I did this here now I could resolve that issue on the C++ SDK

graebm: nevermind. rolling back. Dmitriy is also working on related fixes

DmitriyMusatkin (approved):

PR #745, opened 2025-06-13
remove windows-2019 and latest submodules

TingDaoK:
- Windows 2019 will be fully unsupported by 2025-06-30, https://github.com/actions/runner-images/issues/12045. Remove it with msvc-15 and below.
- Use latest windows 2025 and msvc-17
- Latest submodules


make_pr.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python3
+"""
```
DmitriyMusatkin, line 2: unintentional?

TingDaoK: yep... Removed now

PR #783, opened 2025-09-25
CborDecoder initialize m_lastError in constructor

sbaluja:
*Issue #, if available:*

*Description of changes:*
CborDecoder initialize m_lastError in constructor to AWS_ERROR_SUCCESS instead of it holding a garbage value


TingDaoK (commented):
Can you also help to add a quick regression test, like just create the decoder and ask for the last error, it should return UNKNOWN instead of garbage.

PR #868, opened 2026-06-22
IoT Metrics Support

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Add IoT Metrics Support: 

- **Move `AWSIoTMetrics` from private to public** — New struct `AWSIoTMetrics` that contains `LibraryName` and `AddMetadata()` for users/higher-level SDKs to pass custom metrics into MQTT connections. 
- **Metrics encoding system** — Added a private encoder to detect and encodes 11 client features (protocol version, TLS settings, socket type, proxy config, session behavior, etc.) as compact `ID/Value` pairs in the CONNECT packet
- **MQTT5 integration** — Added a way to setup metrics through `WithSdkMetrics()` API on `Mqtt5ClientOptions`; final metrics will be setup at client creation time
- **MQTT3 integration** — Added a way to setup metrics via new optional parameter on `MqttClient::NewConnection()`; metrics will be setup on `Connect()` call. 
- **IoT builder updates** — Both MQTT5 and MQTT3 IoT builders now pass `IoTSDKVersion` metadata down to Mqtt Client. 

**Note:** Regarding the `enableMetrics` / `disableMetrics` naming convention: other AWS SDKs use `disableMetrics` as the variable name. However, since the C++ API has exposed `enableMetricsCollection` since its inception, we chose to keep the `enableMetrics` naming convention for consistency.


include/aws/crt/io/TlsOptions.h
```diff
@@ -348,12 +360,24 @@ namespace Aws
...
+                    int metricsCertificateSource = 0,
+                    aws_tls_versions metricsTlsVersion = AWS_IO_TLS_VER_SYS_DEFAULTS,
+                    aws_tls_cipher_pref metricsCipherPref = AWS_IO_TLS_CIPHER_PREF_SYSTEM_DEFAULT) noexcept;
                 aws_tls_connection_options m_tls_connection_options;
                 aws_allocator *m_allocator;
                 int m_lastError;
                 bool m_isInit;
 
+                // Internal metrics tracking fields
+                int m_metricsCertificateSource;
```
sfod: Can we change this field type to `CertificateSource`? Forward-declare it to not expose details. Then you'll be able to get rid of all awkward `static_cast`.

include/aws/crt/mqtt/MqttConnection.h
```diff
@@ -10,6 +10,7 @@
 #include <aws/crt/http/HttpConnection.h>
 #include <aws/crt/io/SocketOptions.h>
 #include <aws/crt/io/TlsOptions.h>
+#include <aws/crt/mqtt/IoTSDKMetrics.h>
```
sfod: Is this include needed?

include/aws/crt/mqtt/MqttTypes.h
```diff
@@ -34,6 +35,7 @@ namespace Aws
                 bool useWebsocket = false;
                 bool useTls = false;
                 bool enableMetrics = true;
+                Crt::Optional<IoTDeviceSDKMetrics> sdkMetrics;
                 Allocator *allocator = nullptr;
```
sfod: This breaks source backward compatibility. Though from what I see, there is no way this struct can be used externally even though it's exposed.
Maybe add a comment that this struct is internal?

include/aws/crt/io/TlsOptions.h
```diff
@@ -396,6 +420,10 @@ namespace Aws
 
                 std::shared_ptr<aws_tls_ctx> m_ctx;
                 int m_initializationError;
+                // Internal metrics tracking fields, track the certificate source
+                int m_metricsCertificateSource;
```
sfod: The same comment about changing the type to `CertificateSource`.

include/aws/crt/mqtt/private/IoTSDKMetricsPrivate.h
```diff
@@ -0,0 +1,224 @@
...
+                // Returns '\0' for AWS_IO_TLS_CIPHER_PREF_SYSTEM_DEFAULT (omit from encoded list).
+                static char metricsValueForTlsCipherPreference(aws_tls_cipher_pref pref);
+
+                // Maps aws_tls_versions to its metrics value char.
+                // Returns '\0' for AWS_IO_TLS_VER_SYS_DEFAULTS (omit from encoded list).
+                static char metricsValueForMinimumTlsVersion(aws_tls_versions version);
+
+                // Returns the metrics value char for the socket implementation on the current platform.
+                // Detected at compile time: Winsock on Windows, Apple Network Framework on Apple, POSIX elsewhere.
+                static char detectSocketImplementation();
```
sfod, line 227: With s2n on macos released, we should revisit this function.

include/aws/crt/mqtt/private/IoTSDKMetricsPrivate.h
```diff
@@ -0,0 +1,224 @@
+#pragma once
```
sfod, line 6: We have `/*! \cond DOXYGEN_PRIVATE` in MqttConnectionCore.h. Should we add the same here to not pollute docs?

source/mqtt/IoTSDKMetrics.cpp
```diff
@@ -0,0 +1,477 @@
...
+                return MetricsSocketImplementationValue::Posix;
+#endif
+            }
+
+            void IoTSDKMetricsEncoder::appendFeature(Crt::String &featureList, char featureId, char value)
+            {
+                if (value == '\0')
+                    return;
+                if (!featureList.empty())
+                    featureList += ',';
```
sfod, line 564: nit: let's always use braces

source/mqtt/IoTSDKMetrics.cpp
```diff
@@ -0,0 +1,477 @@
...
+                // Parse CRT features first, then user features (user takes precedence)
+                parseFeatureList(crtFeatures);
+                parseFeatureList(userFeatures);
+
+                // Build merged string
+                Crt::String result;
+                for (const auto &entry : featureMap)
+                {
+                    if (!result.empty())
+                        result += ',';
```
sfod, line 391: nit: let's always use braces

tests/IoTSDKMetricsTest.cpp
```diff
@@ -0,0 +1,545 @@
...
+    MqttClient mqttClient(allocator);
+    ASSERT_TRUE(mqttClient);
+
+    auto connection = mqttClient.NewConnection(
+        "localhost", 8883, socketOptions, false /* no websocket */, true /* enableMetrics */, customMetrics);
+    ASSERT_NOT_NULL(connection.get());
+    ASSERT_TRUE(*connection);
+    return AWS_OP_SUCCESS;
+}
+AWS_TEST_CASE(IoTSDKMetricsMqtt3Minimal, s_TestIoTSDKMetricsMqtt3Minimal)
```
sfod, line 545: Since we parse user-provided feature lists (in `IoTSDKMetricsEncoder::mergeFeatureLists`), I think we should add a few tests for handling invalid data. For example, `A/`, `/,`, etc.

sfod (commented):
Still reviewing, so this is the first part I guess

include/aws/crt/mqtt/IoTSDKMetrics.h
```diff
@@ -0,0 +1,74 @@
...
+
+        namespace Mqtt
+        {
+            /**
+             * IoT Device SDK Metrics Structure.
+             *
+             * Holds the library name and a list of metadata key-value pairs to be appended
+             * to the MQTT CONNECT packet's username field.
+             */
+            struct AWS_CRT_CPP_API IoTDeviceSDKMetrics
```
sfod: This struct looks odd, like it's trying to be both a simple aggregation type and an encapsulated class with invariants, private members, friends.
I think either 1) it should keep only `LibraryName` and `Metadata` with other logic moving to a caller side, or 2) it should become a proper class, with no public member fields.

include/aws/crt/io/TlsOptions.h
```diff
@@ -306,6 +316,8 @@ namespace Aws
              */
             class AWS_CRT_CPP_API TlsConnectionOptions final
             {
+                friend class Mqtt::IoTSDKMetricsEncoder;
```
sfod: Debatable:
I really don't like this - io module should know nothing about mqtt.
As an alternative, which won't require many changes, `TlsConnectionOptions` could expose public member function marked as internal (similar to `GetUnderlyingHandle`), which returns a struct containing fields needed for mqtt metrics. Something like:
```c++
/// @private
TlsConnectionInfo GetTlsConnectionInfo() const noexcept;
```

I'd suggest the same is done for `MqttConnectionCore` and `Mqtt5ClientOptions`. Even though they're in the same mqtt module, adding tight coupling seems unnecessary.

include/aws/crt/io/TlsOptions.h
```diff
@@ -348,12 +360,24 @@ namespace Aws
               private:
                 bool isValid() const noexcept { return m_isInit; }
 
-                TlsConnectionOptions(aws_tls_ctx *ctx, Allocator *allocator) noexcept;
+                // Create TlsConnectionOptions from aws_tls_ctx. The metricsCertificateSource, metricsTlsVersion, and
+                // metricsCipherPref are used for metrics tracking. It might not directly reflect the real options.
+                TlsConnectionOptions(
+                    aws_tls_ctx *ctx,
+                    Allocator *allocator = ApiAllocator(),
```
sfod: Why the allocator parameter became optional?

include/aws/crt/mqtt/MqttTypes.h
xiazhvera: The PR did have some breaking changes. I feel it is fine to break the backward compatibility regardless that the struct could not be used externally?

include/aws/crt/mqtt/private/IoTSDKMetricsPrivate.h
xiazhvera: Updated to use AWS_USE_SECITEM to determin if we enabled nw socket or not.

tests/IoTSDKMetricsTest.cpp
xiazhvera: Updated the function to validate the invalid metrics, and updated the test.

include/aws/crt/io/private/TlsMetrics.h
sfod, line 1: This header file contains not only cert source now, so should be renamed. Maybe `TlsMetrics.h` or something like that?

include/aws/crt/io/TlsOptions.h
```diff
@@ -223,6 +225,11 @@ namespace Aws
 
               private:
                 aws_tls_ctx_options m_options;
+                aws_tls_versions m_tlsVersion;
+                aws_tls_cipher_pref m_cipherPref;
```
sfod: Let's use consistent names. `TlsConnectionOptions` uses different naming:
```cpp
aws_tls_versions m_metricsTlsVersion;
aws_tls_cipher_pref m_metricsCipherPref;
```

include/aws/crt/mqtt/MqttClient.h
```diff
@@ -96,7 +98,9 @@ namespace Aws
                     const Io::SocketOptions &socketOptions,
                     const Crt::Io::TlsContext &tlsContext,
                     bool useWebsocket = false,
-                    bool enableMetrics = true) noexcept;
+                    bool enableMetrics = true,
```
xiazhvera, line 101: Debatable: Regarding the enableMetrics / disableMetrics naming convention: other AWS SDKs use disableMetrics as the variable name. However, since the C++ API has exposed enableMetricsCollection since its inception, we chose to keep the enableMetrics naming convention for consistency.

include/aws/crt/mqtt/IoTSDKMetrics.h
```diff
@@ -0,0 +1,63 @@
...
+
+        namespace Mqtt
+        {
+            /**
+             * IoT Device SDK Metrics Structure.
+             *
+             * Holds the library name and a list of metadata key-value pairs to be appended
+             * to the MQTT CONNECT packet's username field.
+             */
+            struct AWS_CRT_CPP_API AWSIoTMetrics
```
sfod: The naming style here is not well defined, but `AwsIotMetrtics` fits better IMO.

source/mqtt/IoTSDKMetrics.cpp
```diff
@@ -0,0 +1,493 @@
...
+                        // A valid feature token must have the format "X/Y" where:
+                        // - The feature ID is exactly one character (slashPos == 1)
+                        // - The feature ID is a recognized MetricsFeatureId
+                        // - The value (after slash) is non-empty
+                        if (slashPos == 1 && slashPos + 1 < token.size() && isValidFeatureId(token[0]))
+                        {
+                            featureMap[token.substr(0, slashPos)] = token.substr(slashPos + 1);
+                        }
+                        if (commaPos == Crt::String::npos)
+                            break;
```
sfod, line 375: Missing braces.

source/mqtt/IoTSDKMetrics.cpp
sfod: Braces still missed.

source/mqtt/IoTSDKMetrics.cpp
```diff
@@ -0,0 +1,493 @@
...
+            }
+
+            ////////// AWSIoTMetrics //////////
+
+            void AWSIoTMetrics::initializeRawOptions(struct aws_mqtt_iot_metrics &raw_options) noexcept
+            {
+                raw_options.library_name = ByteCursorFromString(libraryName);
+
+                // Rebuild the raw entry array from the current Metadata contents.
+                // Byte cursors point directly into the strings stored in Metadata.
```
sfod: Considering that Metadata is a public member, this is brittle. Also, the copy members generated by a compiler will invalidate pointers stored in `m_rawMetadataEntries`. Currently, there are no issues possible afaics. But having a class with such "surprise" might silently break in the future (e.g. if we add it as a member to another copyable class).
I suggest the following:
1) make `metadata` and `libraryName` private, add public setters where `m_rawMetadataEntries` will be invalidated.
2) Implement copy and move members handling pointers invalidation.

include/aws/crt/mqtt/IoTSDKMetrics.h
xiazhvera: As we have using the name for other SDKs, we will keep it here.

sfod (approved):

PR #871, opened 2026-06-26
Bindings for BDD Endpoint Resolution

azkrishpy:
*Issue #, if available:*

*Description of changes:*
https://github.com/awslabs/aws-c-sdkutils/pull/62
Bindings for BDD Endpoint Resolution.

Also, unified tests for legacy and the bdd engine.


include/aws/crt/endpoints/BddEngine.h
```diff
@@ -0,0 +1,79 @@
...
+                    Allocator *allocator,
+                    const ByteBuf &bytecodeBuffer,
+                    const ByteCursor &partitionsCursor) noexcept;
+
+                /**
+                 * Construct from a bytecode cursor.
+                 * The caller is responsible for keeping the underlying buffer alive
+                 * for the lifetime of this engine.
+                 */
+                BddEngine(
```
sbiscigl, line 36: `s_initEngine` takes a byte cursor, seems like we should only really need this constructor. seems like it would be a good idea just to have only this one constructor. in fact in this constructor `m_bytecodeBuf` is not used at all and you can completely get rid of.

include/aws/crt/endpoints/BddEngine.h
```diff
@@ -0,0 +1,79 @@
...
+
+                /**
+                 * Resolve an endpoint from the provided request context.
+                 * Returns None on failure; use Aws::Crt::LastError() to retrieve the error code.
+                 */
+                Optional<ResolutionOutcome> Resolve(const RequestContext &context) const;
+
+              private:
+                ByteBuf m_bytecodeBuf;
+                aws_endpoints_bdd_engine *m_engine;
```
sbiscigl: dont use naked c pointers, use `ScopedResource` like how its done in in XXHash.
header - https://github.com/awslabs/aws-crt-cpp/blob/d35caf1c4e71482843c4c310c290300184b9488a/include/aws/crt/checksum/XXHash.h#L89
source - https://github.com/awslabs/aws-crt-cpp/blob/main/source/checksum/XXHash.cpp#L30

include/aws/crt/endpoints/BddEngine.h
azkrishpy: addressed both comments, thanks!

include/aws/crt/endpoints/BddEngine.h
```diff
@@ -0,0 +1,64 @@
...
+            class AWS_CRT_CPP_API BddEngine final
+            {
+              public:
+                /**
+                 * Construct from a bytecode cursor and partitions JSON.
+                 * The caller is responsible for keeping the bytecode buffer alive
+                 * for the lifetime of this engine.
+                 */
+                BddEngine(
+                    Allocator *allocator,
```
sbiscigl: lets default the allocator the globally registered one to have it match other places in the CRT, this simplfies it for consumers to not add that extra arg. i.e. like how you do it in [URI](https://github.com/awslabs/aws-crt-cpp/blob/2d48365a672944b205ac1a10c74f25cd0d8e5bec/include/aws/crt/io/Uri.h#L29) so this effectively becomes

```c++
BddEngine(
  const ByteCursor &bytecodeCursor,
  const ByteCursor &partitionsCursor,
  Allocator *allocator = ApiAllocator()) noexcept;
```

source/endpoints/BddEngine.cpp
```diff
@@ -0,0 +1,42 @@
...
+    {
+        namespace Endpoints
+        {
+            BddEngine::BddEngine(
+                Allocator *allocator,
+                const ByteCursor &bytecodeCursor,
+                const ByteCursor &partitionsCursor) noexcept
+                : m_engine(nullptr, aws_endpoints_bdd_engine_release)
+            {
+                auto partitions = aws_partitions_config_new_from_string(allocator, partitionsCursor);
```
sbiscigl, line 21: does partitions config need to outlive `m_engine`, or are partitions copied into mem? would it make sense to have two scoped resources, one for partitions, one for rule engine? i.e.
```c++
class BddEngine {
public:
  BddEngine(...);
private:
  ScopedResource<struct aws_partitions_config> m_partitions;
  ScopedResource<struct aws_endpoints_bdd_engine> m_engine;
}
```

then you could build it super easily like
```c++
BddEngine::BddEngine(...)
  : m_partitions(aws_partitions_config_new_from_string(allocator, partitionsCursor), aws_partitions_config_release)
    m_engine(m_partitions, aws_endpoints_bdd_engine_release)
```

without any logic in the body

azkrishpy: the parititions config has not changed in structure so we have reused the implementation from the previous engine. It does not need to be outliving the engine because the values are internally copied.

include/aws/crt/endpoints/BddEngine.h
azkrishpy: fixed, thanks!

source/endpoints/BddEngine.cpp
DmitriyMusatkin: i think for performance reasons we want to avoid copying data around as much as possible. partitions in general copies some data and might be currently safe, but i dont think its assumption that we should rely on. 
at the timing scales we are talking about, copying of data actually adds dramatical overhead to how fast resolution is done (ideally, on sdk side endpoints and partition data just live in the text section and engine just points to that)

tests/resources/sample_partitions.json
```diff
@@ -0,0 +1,45 @@
+{
```
DmitriyMusatkin, line 1: not necessary for this pr, but we should probably unify the sample data used for both old and new engines

DmitriyMusatkin:
srcript to encode ruleset into bdd form is in c-sdkutils, would cpp be able to use it if they only rely on bindings?

DmitriyMusatkin (approved):

PR #889, opened 2026-07-22
Use AWS_WARNINGS_ARE_ERRORS to restrict -Werror in elasticurl_cpp.

jkroonza:
Found by Gentoo tinderbox: https://bugs.gentoo.org/977183

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #891, opened 2026-07-24
Use AWS_WARNINGS_ARE_ERRORS to restrict -Werror

xiazhvera:
This is in some of the sub folders.

Found by Gentoo tinderbox: https://bugs.gentoo.org/977183

*Issue #, if available:*
Bring in pull request: https://github.com/awslabs/aws-crt-cpp/pull/889

*Description of changes:*


bin/elasticurl_cpp/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ if (MSVC)
         target_compile_options(${ELASTICURL_CPP_PROJECT_NAME} PRIVATE "/MD$<$<CONFIG:Debug>:d>")
     endif()
     target_compile_options(${ELASTICURL_CPP_PROJECT_NAME} PRIVATE /W4 /WX)
-else ()
+elseif (AWS_WARNINGS_ARE_ERRORS)
     target_compile_options(${ELASTICURL_CPP_PROJECT_NAME} PRIVATE -Wall -Wno-long-long -pedantic -Werror)
```
sfod: Even though no one complained about Windows, it has the same problem - `/W4 /WX` added by default. Let's change this logic to how it's implemented in other places:
```cmake
# set extra warning flags
if(AWS_WARNINGS_ARE_ERRORS)
    if(MSVC)
        target_compile_options(${TEST_BINARY_NAME} PRIVATE /W4 /WX)
    else()
        target_compile_options(${TEST_BINARY_NAME} PRIVATE -Wall -Wno-long-long -pedantic -Werror)
    endif()
endif()
```

And the same for two other apps.

sfod (approved):

sfod:
`android-armv7` is failing because of the broken podman in the new version of GH Ubuntu 24.04.
See https://github.com/actions/runner-images/issues/14473

PR #893, opened 2026-07-27
Add tcp nodelay option to socket options

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/io/SocketOptions.cpp
+++ b/source/io/SocketOptions.cpp
@@ -23,6 +23,7 @@ namespace Aws
                 options.keep_alive_timeout_sec = 0;
                 options.keep_alive_interval_sec = 0;
                 options.keepalive = false;
+                options.tcp_nodelay = AWS_SOCKET_TCP_NODELAY_ON;
             }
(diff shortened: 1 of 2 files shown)


source/io/SocketOptions.cpp
```diff
@@ -23,6 +23,7 @@ namespace Aws
                 options.keep_alive_timeout_sec = 0;
                 options.keep_alive_interval_sec = 0;
                 options.keepalive = false;
+                options.tcp_nodelay = AWS_SOCKET_TCP_NODELAY_ON;
```
azkrishpy, line 26: on by default? I thought we were doing that only for s3?

azkrishpy (approved):

source/io/SocketOptions.cpp
DmitriyMusatkin: yeah. had a lot of back and forth with java and cpp. decided to have it default on in bindings. all java http clients and curl have it on since forever, so we think risks are minimal

PR #895, opened 2026-07-29
Use latest version of aws-crt-builder

sfod:
Update aws-crt-builder version to `latest`.


xiazhvera (approved):
Trivial, there is an EOF change.

PR #899, opened 2026-07-31
Updates to CRT-CPP Bindings for TM upload/download

narahavi:
*Issue #, if available:*

*Description of changes:*

Updating the bindings based on updates to the single file upload/download functionality of TM 2.0 in aws-sdk-cpp

### DIFF
--- a/include/aws/crt/s3/S3.h
+++ b/include/aws/crt/s3/S3.h
@@ -221,6 +221,54 @@ namespace Aws
             class S3MetaRequestOptions;
 
             /**
+             * C++ mirror of the aws-c-s3 error codes reported on
+             * S3MetaRequestResult, so callers can switch on a typed enum instead
+             * of pulling in the C header for AWS_ERROR_S3_*. Success indicates
+             * the meta request completed without error; Unknown is returned for
+             * any code the binding does not (yet) recognize.
+             */
+            enum class S3ErrorCode
+            {
+                Success,
+                Unknown,
+                Canceled,
+                Paused,
+                SlowDown,
+                RequestTimeout,
+                RequestTimeTooSkewed,
+                InvalidResponseStatus,
+                ResponseChecksumMismatch,
+                ChecksumCalculationFailed,
+                IncorrectContentLength,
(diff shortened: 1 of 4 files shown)


source/s3/S3.cpp
```diff
@@ -316,6 +415,44 @@ namespace Aws
...
+                {
+                    m_lastError = AWS_ERROR_INVALID_ARGUMENT;
+                    return *this;
+                }
+
+                // Zeroed so a re-set leaves no stale cursors.
+                aws_signing_config_aws &config = m_impl->endpointSigningConfig;
+                AWS_ZERO_STRUCT(config);
+                aws_s3_init_default_signing_config(
+                    &config, ByteCursorFromString(region), provider->GetUnderlyingHandle());
```
sbaluja, line 361: Is it guaranteed that the provider pointer will be alive when we call `MakeMetaRequest`, because at that point the signingConfig will access the reference we are storing

sbaluja (approved):

sbiscigl (approved):

include/aws/crt/s3/S3.h
```diff
@@ -232,6 +280,14 @@ namespace Aws
                 /** AWS_ERROR_SUCCESS on success, a CRT error code otherwise. */
                 int errorCode;
 
+                /**
+                 * Typed view of errorCode. Returns S3ErrorCode::Success when errorCode
+                 * is AWS_ERROR_SUCCESS, S3ErrorCode::Unknown for any code the binding
+                 * does not recognize. Lets callers switch on the enum without pulling
+                 * in the aws-c-s3 C header.
```
DmitriyMusatkin: dont know how i feel about redefining all enum values in a new header. would it be simpler to just include c-s3 in your header and let users just use original values? maybe even namespace them

source/Api.cpp
```diff
@@ -461,6 +461,11 @@ namespace Aws
             return aws_error_debug_str(error);
         }
 
+        const char *ErrorString(int error) noexcept
```
sbaluja: Call this func ErrorName

PR #903, opened 2026-08-21
Fix aws_get_version conflict

azkrishpy:
*Issue #, if available:*

*Description of changes:*
There is a conflict in naming with aws-c-common helper. This PR should fix that.


cmake/AwsGetFullVersion.cmake
```diff
@@ -1,4 +1,4 @@
-function(aws_get_version var_version_simple var_version_full var_git_hash)
+function(aws_get_full_version var_version_simple var_version_full var_git_hash)
```
TingDaoK: if it conflict with what we have in aws-c-common, should we name it as aws-crt-cpp-get-version or something?

azkrishpy: sure I can change it, it is just that I thought functions like this would also go into common eventually instead of each C adjacent lib maintaining its own cmake so made the name more generic..

TingDaoK (approved):

